PR #26, opened 2020-01-16
Connect to artifact repositories using HTTPS

rschev:
See:
- https://support.sonatype.com/hc/en-us/articles/360041287334
- https://jfrog.com/blog/secure-jcenter-with-https/

Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-404

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,6 @@ def getVersionQualifier(String branch_name) {
 
 buildscript {
     repositories {
-        jcenter()
         mavenCentral()
         maven {
             url "https://maven.alfresco.com/nexus/content/groups/public"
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -77,10 +77,10 @@ subprojects {
 
     repositories {
         mavenLocal()
-        maven { url "http://repo.maven.apache.org/maven2" }
+        mavenCentral()
+        jcenter()
```
kerkhofsd: Do you need `jcenter()`?
It might have been required because in the past DE was published to jcenter. 

Just checking, if it's required for other dependencies as well no problem.

rschev: Good call, it seems to build just fine without it.

PR #31, opened 2020-02-18
ALFREDAPI-410 wrap Data Dictionary lookup with runAsSystem

thijslemmens:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-410

- [ X] Is [CHANGELOG.md](https://github.com/xenit-eu/alfred-api/blob/master/CHANGELOG.md) extended?
- [ X] Does this PR avoid breaking the API? 


apix-impl/src/main/java/eu/xenit/apix/alfresco/configuration/ConfigurationServiceImpl.java
```diff
@@ -60,7 +61,7 @@ public Configurations getConfigurationFiles(String searchDirectory, String nameF
         }
         String[] searchDirectoryParts = searchDirectory.split("/");
 
-        NodeRef rootFolder = fileFolderService.getDataDictionary();
+        NodeRef rootFolder = AuthenticationUtil.runAsSystem(() -> fileFolderService.getDataDictionary());
```
WimCrols: I would like a comment like _Running as system so user does not need read access on the folders inbetween_

thijslemmens: et voilà

PR #36, opened 2020-04-09
Ha 29

kpiot123:
Fixes https://xenitsupport.jira.com/browse/HA-29


WimCrols:
In Alfred API we only put tickets from project "ALFREDAPI". I know it's a bit of a hassle but can you make products tickets there, and reference those in your commits?

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -482,6 +486,57 @@ public void getAllInfoOfNodes(WebScriptRequest request, WebScriptResponse respon
...
+                    + "\n"
+                    + "Parameter 'rootRef' is an optional parameter. If the parameter is not given the default value\n"
+                    + "will be the node reference of Company Home. It is the last parent last parent in the list of\n"
+                    + "recursive parents."
+                    + "Parameter 'nodeRefs' is a mandatory parameter. It contains the node references of the nodes\n"
+                    + "for which the recursive parents need to be retrieved.\n"
+                    + "\n"
+                    + "Nodes for which the recursive parents are requested are omitted from the result if the node\n"
+                    + "does not exist or one of its recursive parents cannot be read by the user.")
+    @Uri(value = "/nodes/parent/recursive", method = HttpMethod.POST)
```
WimCrols, line 509: As discussed over Slack, I would like a single-object operation via HTTP `GET`. The overhead of multiple calls can be avoided via the bulk call .

apix-impl/src/main/java/eu/xenit/apix/alfresco/sites/SiteService.java
```diff
@@ -0,0 +1,73 @@
...
+    private final static String DOCUMENT_LIBRARY_COMPONENT = "documentLibrary";
+    private final static String LINKS_COMPONENT = "links";
+    private final static String DATA_LISTS_COMPONENT = "dataLists";
+    private final static String WIKI_COMPONENT = "wiki";
+    private final static String DISCUSSIONS_COMPONENT = "discussions";
+
+    private ApixToAlfrescoConversion c;
+
+    @Autowired
+    private ServiceRegistry serviceRegistry;
```
WimCrols, line 33: Both field **and** constructor injection are being used here. I don't think that is right.
I know APIX is a bit of a mess when it comes to injection, but since this is a new class, let's use 100% constructor injection.

apix-impl/src/test/java/eu/xenit/apix/tests/metadata/RecursiveParentsBaseUnitTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import eu.xenit.apix.alfresco.ApixToAlfrescoConversion;
+import org.alfresco.service.ServiceRegistry;
+import org.alfresco.service.cmr.repository.NodeRef;
+import org.alfresco.service.cmr.repository.NodeService;
+import org.alfresco.service.cmr.security.AccessStatus;
+import org.alfresco.service.cmr.security.PermissionService;
+import org.junit.Before;
+
+public abstract class RecursiveParentsBaseUnitTest {
```
WimCrols, line 15: Thanks for making the effort for making this run as unit test and not integration test :+1:

apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
```diff
@@ -421,6 +426,56 @@ public NodeAssociations getAssociations(eu.xenit.apix.data.NodeRef ref) {
         return new NodeAssociations(getChildAssociations(ref), getParentAssociations(ref), getTargetAssociations(ref));
     }
 
+    @Override
+    public List<eu.xenit.apix.data.NodeRef> getRecursiveParents(eu.xenit.apix.data.NodeRef ref, eu.xenit.apix.data.NodeRef rootRef) {
```
WimCrols, line 430: Please call these methods `getAncestors` (etc)

WimCrols: Is there ever a use case where `rootRef` is **not** company home? When you know your e.g. grand child node, and you know your grandparent node, don't you also know all the ancestors inbetween?

apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
```diff
@@ -421,6 +426,56 @@ public NodeAssociations getAssociations(eu.xenit.apix.data.NodeRef ref) {
...
+            ChildAssociationRef parentAssoc = nodeService.getPrimaryParent(parentRef);
+            if (parentAssoc == null) return null;
+            parentRef = parentAssoc.getParentRef();
+        }
+
+        return recursiveParentRefs;
+    }
+
+    @Override
+    public Map<eu.xenit.apix.data.NodeRef, List<eu.xenit.apix.data.NodeRef>> getRecursiveParents(
```
WimCrols, line 466: Since I give the suggestion to go for single-object calls, this function can be removed.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -482,6 +486,57 @@ public void getAllInfoOfNodes(WebScriptRequest request, WebScriptResponse respon
         //logger.error("end writeJsonResponse");
     }
 
+    @ApiOperation(value = "Retrieves the parents of the nodes recursively",
+            notes = "example json request data: \n"
```
WimCrols, line 490: Thanks for the example doc :+1:

WimCrols: Don't forget to update it when you update method names

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -482,6 +486,57 @@ public void getAllInfoOfNodes(WebScriptRequest request, WebScriptResponse respon
...
+            throw new InvalidArgumentException("nodeRefsJsonArray is null");
+        }
+
+        NodeRef rootRef = null;
+        if (jsonObject.has("rootRef")) {
+            rootRef = new NodeRef(jsonObject.getString("rootRef"));
+            logger.debug("Set rootRef to " + rootRef);
+        }
+
+        int nodeRefsJsonArrayLength = nodeRefsJsonArray.length();
```
WimCrols, line 528: (Minor detail) iteration is not possible with JsonArays?

apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
kpiot123: In Alfred Desktop it is possible to change the root to a lower level. For example you only want users to be able to use sites (which is a child of company home). 
I can also imagine that for performance reasons you may not want to return all ancestors if you already know the ancestors up to a certain node and you don't want to retrieve them again.
It may also be used to check if a node is the child of another node. For example, I set the rootRef to the node reference of Shared and in the result the node reference of the last ancestor is not Shared then i know that a node is not a child of Shared.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
kpiot123: no, I can't get an iterator from it.

apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
WimCrols: Good, sounds like it is a use case indeed.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
WimCrols: Ok np

WimCrols:
Closed because replaced by https://github.com/xenit-eu/alfred-api/pull/37

PR #37, opened 2020-04-17
ALFREDAPI-417

kpiot123:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-417


kpiot123:
This is the updated version of pull request https://github.com/xenit-eu/alfred-api/pull/36. This one has the correct branch name.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/ApixV1Webscript.java
```diff
@@ -81,6 +82,24 @@ protected NodeInfo nodeRefToNodeInfo(NodeRef nodeRef, IFileFolderService fileFol
         return nodeInfoList;
     }
 
+    protected NodeInfo nodeRefToNodeInfo(NodeRef nodeRef,
```
WimCrols, line 85: If you have 2 conversion methods --one for a single object and one for lists-- then the list one should use the single one, not vice versa IMO.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/sites/SitesWebscript1.java
```diff
@@ -0,0 +1,117 @@
...
+    public void getMySites(@RequestParam(required = false) Boolean retrieveMetadata,
+            @RequestParam(required = false) Boolean retrievePath,
+            @RequestParam(required = false) Boolean retrievePermissions,
+            @RequestParam(required = false) Boolean retrieveChildAssocs,
+            @RequestParam(required = false) Boolean retrieveParentAssocs,
+            @RequestParam(required = false) Boolean retrieveTargetAssocs,
+            WebScriptResponse response)
+            throws IOException {
+        logger.debug("retrieveMetadata: "+retrieveMetadata);
+        logger.debug("retrievePath: "+retrievePath);
```
WimCrols, line 78: Please reformat

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/ApixV1Webscript.java
kpiot123: The way I see it a single object is a special case of multiple objects. I know that in some cases following this philosophy can be more efficient. For example if you have a method that handles one object and does some network calls. If you have to do this for multiple objects it would mean you have to do those calls also multiple times because you loop over the same method multiple times. So the more objects you have the more network calls are needed. Instead, if you would handle the single object as if it were multiple objects then you could in theory always do this in the same amount of calls. This is just an example but I think it illustrates the difference.

Do you have other reasons why a method for multiple objects should use the one for a single object?

WimCrols: From an **API perspective** I agree with you, and it is a good principle to follow. But this is not an API, but internal code structure (so network overhead is not relevant). 

The usual way of writing code is *loop conversion > foreach of single conversion*. This is *single conversion > transform to list -> loop conversion -> transform back to single*
Not only does that deviate from what most readers would expect, it introduces unnecessary complexity of 2 conversions.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ## 2.4.1 (Unreleased)
```
koenvangeerteruy, line 4: I believe the version needs to be updated from 2.4.1 to 2.5.0 since we are adding new functionality following guidelines in XEP-7 and semantic versioning we are trying to follow for Alfred API

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/sites/SitesWebscript1.java
```diff
@@ -0,0 +1,117 @@
...
+        logger.debug("retrievePermissions: "+retrievePermissions);
+        logger.debug("retrieveChildAssocs: "+retrieveChildAssocs);
+        logger.debug("retrieveParentAssocs: "+retrieveParentAssocs);
+        logger.debug("retrieveTargetAssocs: "+retrieveTargetAssocs);
+
+        if (retrieveMetadata == null){
```
koenvangeerteruy, line 84: I would suggest to set the property defaultValue in @RequestParam (https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestParam.html) instead of writing null checks for each optional boolean request parameter.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/ApixV1Webscript.java
koenvangeerteruy: @WimCrols I think it depends on the point of view and approach. I think there are 2 options:
1. Collection conversion is foreach of single conversion
2. Single conversion is a special case of collection conversion
Both are valid options and they have both pros and cons. Depending on use case, complexity of processing a single conversion and  performance overhead/considerations one of the options is preferable.

I would follow the general convention used in this project for implementing methods covering both single and collection arguments.

PR #39, opened 2020-05-12
ALFREDAPI-416 Removed max repository version from module.properties for builds for most recent Alfresco 

kpiot123:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-416

### DIFF
--- a/apix-impl/52/src/main/config/module.properties
+++ b/apix-impl/52/src/main/config/module.properties
@@ -4,4 +4,4 @@ module.description=${project.description}
 module.version=${project.version}
 
 module.repo.version.min=${project.alfresco_min_version}
-module.repo.version.max=${project.properties["alfresco_max_version"]}
+module.repo.version.max=${project.properties.alfresco_max_version}
--- a/apix-impl/52/src/main/config/module.properties
+++ b/apix-impl/52/src/main/config/module.properties
@@ -3,5 +3,5 @@ module.title=${project.name}
 module.description=${project.description}
 module.version=${project.version}
 
-module.repo.version.min=5.2
-module.repo.version.max=5.2.99
+module.repo.version.min=${project.alfresco_min_version}
+module.repo.version.max=${project.properties["alfresco_max_version"]}
--- a/apix-impl/52/src/main/config/module.properties
+++ b/apix-impl/52/src/main/config/module.properties
@@ -4,4 +4,4 @@ module.description=${project.description}
 module.version=${project.version}
 
 module.repo.version.min=${project.alfresco_min_version}
-module.repo.version.max=${project.properties.alfresco_max_version}
+module.repo.version.max=${project.alfresco_max_version}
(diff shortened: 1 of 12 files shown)


kpiot123:
Does this change also require an update of the changelog? Functionally nothing really changed.

koenvangeerteruy:
@kpiot123 Changelog update? I think this change involves a change that can be of importance for end users.

kpiot123:
> @kpiot123 Changelog update? I think this change involves a change that can be of importance for end users.

I wasn't sure but it can't hurt to add it. I have updated it.

apix-impl/52/src/main/config/module.properties
```diff
@@ -3,5 +3,5 @@ module.title=${project.name}
 module.description=${project.description}
 module.version=${project.version}
 
-module.repo.version.min=5.2
-module.repo.version.max=5.2.99
+module.repo.version.min=${project.alfresco_min_version}
+module.repo.version.max=${project.properties["alfresco_max_version"]}
```
WimCrols, line 7: Why not `${project.properties.alfresco_max_version}`?

apix-impl/build.gradle
```diff
@@ -4,6 +4,31 @@ plugins {
...
+
+    if (currentProject.getName().equals(subProjectList.last().getName())) {
+        return ""
+    }
+
+    return minAlfrescoVersion + ".99"
+}
+
+ext {
+    alfresco_50_min_version = "5.0"
```
WimCrols, line 25: The versioning is already spread out over 5 files, so I prefer not to add another one. Can you remove these constants and simply hardcode the versions in the respective overrides.gradle?

koenvangeerteruy: OK. In that case we can also just remove or set the max_version in overrides.gradle to what it needs to be.
Unless, you want setting the max version to be automated when support for a new Alfresco version is added

kpiot123: To be clear, can the getMaxRepositoryVersionRequirement method be kept or should I also hard-code the max Alfresco version?

WimCrols: That's also ok for me.

WimCrols: To answer Koen: no, no automation is needed for that.

koenvangeerteruy: @WimCrols OK. So we set (hard-code) min and max Alfresco version in the override.gradle files?

WimCrols: Yes

apix-impl/61/overrides.gradle
```diff
@@ -3,4 +3,5 @@ description = "Xenit API-X implementation Alfresco 6.1"
 ext {
     alfresco_version = alfresco_61_version
     alfresco_dm_version = alfresco_61_dm_version
+    alfresco_min_version = "6.1"
```
WimCrols, line 6: @kpiot123 could you add the comment here `Not setting max version to make it easier to test on next Alfresco version (only for last version)`, please? I will already approve. No need to assign back to me. After adding this, you may merge.

PR #44, opened 2020-06-24
Replace terms with property values that will cause Invalid Number Exceptions in Solr 6 with dummy terms in FtsNodeVisitor

### LINKED ISSUE #42: Search: Property values that can never match should be filtered out
It is currently possible to search for a property, with a value that is impossible to exist, due to type definitions in Alfresco. An easy example which can be used to reproduce this, is the `sys:node-dbid` property, which is a property of type `d:long`. Searching for that property with a value of `Long.MAX_VALUE + 1` will return an error. Searching with `Long.MAX_VALUE` succeeds and just returns an empty set (as no one has that many nodes).

```curl -v -u admin:admin localhost:32787/alfresco/s/apix/v1/search -XPOST -H 'Content-Type: application/json' -d '{"query":{"and":[{"or":[{"property":{"name":"sys:node-dbid","value":"9223372036854775808"}}]}]}}'```

Systems that query Alfresco (through Alfred Api for example) should not need to know about the datatype of properties that are in the custom models of the Alfresco. In our case, a system (Alfred Finder) expands a value to all sets of property name : that value
As a necessary workaround, Alfred Api can check the datatype using the Alfresco `DictionaryService` and filter out the terms that do not comply with the datatype of their property value, before sending out the query to Alfresco.

### PR DESCRIPTION
**Pull request by driesdeproost**
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-427
Fixes #42 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 ### Fixed
 * [ALFREDAPI-419](https://xenitsupport.jira.com/browse/ALFREDAPI-419): Add urldecoding to deleteAssociation endpoint
 * [ALFREDAPI-422](https://xenitsupport.jira.com/browse/ALFREDAPI-422): Fix totalResults and limits for TDMQ's
+* [ALFREDAPI-427](https://xenitsupport.jira.com/browse/ALFREDAPI-427): Replace terms with property values that will cause Invalid Number Exceptions in Solr 6 with dummy terms in FtsNodeVisitor
+
 ### Deleted
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 ### Fixed
 * [ALFREDAPI-419](https://xenitsupport.jira.com/browse/ALFREDAPI-419): Add urldecoding to deleteAssociation endpoint
 * [ALFREDAPI-422](https://xenitsupport.jira.com/browse/ALFREDAPI-422): Fix totalResults and limits for TDMQ's
-* [ALFREDAPI-427](https://xenitsupport.jira.com/browse/ALFREDAPI-427): Replace terms with property values that will cause Invalid Number Exceptions in Solr 6 with dummy terms in FtsNodeVisitor
+* [ALFREDAPI-427](https://xenitsupport.jira.com/browse/ALFREDAPI-427): Add workaround to prevent Solr Exceptions when searching for numeric values that overflow when parsed as int or long
 
 ### Deleted
 
--- a/apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
+++ b/apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
@@ -34,7 +34,6 @@ import org.alfresco.service.transaction.TransactionService;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -321,5 +320,32 @@ abstract public class SearchServiceTest extends BaseTest {
      * Implemented in alfresco version specific subclasses to set up and tear down solr facets.
      */
     abstract protected void withTestFacets(AuthenticationUtil.RunAsWork<Object> work) throws Exception;
+
+
+    @Test
+    public void testNotFilteredOut() {
+        SearchSyntaxNode node =
+                new QueryBuilder()
+                        .startOr()
+                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", "9223372036854775807")
+                        .end()
+                        .create();
+        SearchQuery searchQuery = new SearchQuery();
+        searchQuery.setQuery(node);
+        searchService.query(searchQuery);
+    }
+
+    @Test
+    public void testNothingLeftAfterFiltering() {
+        SearchSyntaxNode node =
+                new QueryBuilder()
+                        .startOr()
+                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", "9223372036854775808")
+                        .end()
+                        .create();
+        SearchQuery searchQuery = new SearchQuery();
+        searchQuery.setQuery(node);
+        searchService.query(searchQuery);
+    }
 
 }
--- a/apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
+++ b/apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
@@ -14,6 +14,7 @@ import eu.xenit.apix.tests.BaseTest;
 import eu.xenit.apix.util.SolrTestHelper;
 import java.io.IOException;
 import java.io.Serializable;
+import java.math.BigInteger;
 import java.util.Locale;
 import java.util.Map;
 import javax.sql.DataSource;
@@ -42,6 +43,10 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.extensions.surf.util.I18NUtil;
 
 abstract public class SearchServiceTest extends BaseTest {
+
+    private static final String LONG_MAX_VALUE = String.valueOf(Long.MAX_VALUE);
+    private static final String LONG_MAX_VALUE_PLUS_ONE = new BigInteger(LONG_MAX_VALUE).add(new BigInteger("1"))
+            .toString();
 
     private final static Logger logger = LoggerFactory.getLogger(SearchServiceTest.class);
     private static final String ADMIN_USER_NAME = "admin";
@@ -323,11 +328,11 @@ abstract public class SearchServiceTest extends BaseTest {
 
 
     @Test
-    public void testNotFilteredOut() {
+    public void validLong_onlyNode_doesNotThrowException() {
         SearchSyntaxNode node =
                 new QueryBuilder()
                         .startOr()
-                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", "9223372036854775807")
+                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", LONG_MAX_VALUE)
                         .end()
                         .create();
         SearchQuery searchQuery = new SearchQuery();
@@ -336,11 +341,11 @@ abstract public class SearchServiceTest extends BaseTest {
     }
 
     @Test
-    public void testNothingLeftAfterFiltering() {
+    public void invalidLong_onlyNode_doesNotThrowException() {
         SearchSyntaxNode node =
                 new QueryBuilder()
                         .startOr()
-                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", "9223372036854775808")
+                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", LONG_MAX_VALUE_PLUS_ONE)
                         .end()
                         .create();
         SearchQuery searchQuery = new SearchQuery();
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -13,6 +13,8 @@
 ### Fixed
 * [ALFREDAPI-419](https://xenitsupport.jira.com/browse/ALFREDAPI-419): Add urldecoding to deleteAssociation endpoint
 * [ALFREDAPI-422](https://xenitsupport.jira.com/browse/ALFREDAPI-422): Fix totalResults and limits for TDMQ's
+* [ALFREDAPI-427](https://xenitsupport.jira.com/browse/ALFREDAPI-427): Replace terms with property values that will cause Invalid Number Exceptions in Solr 6 with dummy terms in FtsNodeVisitor
```
tgeens, line 16: @driesdeproost the changelog items should state what issue this solves (audience:user), not how (audience: developer)

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
```diff
@@ -322,4 +321,31 @@ private void withLocale(Locale locale, AuthenticationUtil.RunAsWork<Object> work
...
+    public void testNotFilteredOut() {
+        SearchSyntaxNode node =
+                new QueryBuilder()
+                        .startOr()
+                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", "9223372036854775807")
+                        .end()
+                        .create();
+        SearchQuery searchQuery = new SearchQuery();
+        searchQuery.setQuery(node);
+        searchService.query(searchQuery);
```
tgeens, line 335: what is being tested here ? if it is "does not throw exception", either:
- put that in the method name
- assert that the return value is not null

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
```diff
@@ -322,4 +321,31 @@ private void withLocale(Locale locale, AuthenticationUtil.RunAsWork<Object> work
...
+    }
+
+    @Test
+    public void testNothingLeftAfterFiltering() {
```
tgeens, line 339: this is EXACTLY the same test as above ?

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/PropertyTypeCheckService.java
```diff
@@ -0,0 +1,68 @@
...
+            logger.warn("Instantiating PropertyTypeCheckService without a PropertyService, type checks will always succeed");
+        }
+        this.propertyService = propertyService;
+    }
+
+    public boolean fitsType(PropertySearchNode node) {
+        if (node == null || node.getValue() == null || node.getName() == null || propertyService == null) {
+            return true;
+        }
+        QName qName = new QName(node.getName().replaceAll("\\\\", "")); //stuff like hyphen in node-dbid is escaped
```
tgeens, line 37: isn't there an API-way to do this ? feels like we're missing potential edge cases here ?

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
driesdeproost: exactly, does not throw exception
agreed

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
driesdeproost: It might seem so, but the long value is 1 greater, crossing the max value boundary

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/PropertyTypeCheckService.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Map.Entry;
+import java.util.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PropertyTypeCheckService {
+
+    private final static Logger logger = LoggerFactory.getLogger(PropertyTypeCheckService.class);
+
+    private final Map<String, Function<String, Boolean>> constraintMap = new HashMap<String, Function<String, Boolean>>() {{
```
tgeens, line 17: constraint is an overloaded term here
Suggestions: `dataTypePredicates`

tgeens, line 17: Replace `Function<String, Boolean>` with `Predicate<String>` ?

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/PropertyTypeCheckService.java
driesdeproost: Good question, I will look into it

apix-impl/src/test/java/eu/xenit/apix/tests/search/FtsNodeVisitorTest.java
```diff
@@ -0,0 +1,96 @@
...
+            querySyntaxTree = querySyntaxTree.property(property, value);
+        }
+        return querySyntaxTree
+                .end()
+                .term("type", "{http://www.alfresco.org/model/content/1.0}content")
+                .end()
+                .create();
+    }
+
+    private PropertyService generatePropertyServiceMock() {
```
tgeens, line 76: suggestion: create a `PropertyServiceStub` (inner?) class instead that accepts that `propertyToDataType` map as constructor arg

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
tgeens: Ok, need to make that more obvious somehow ;-)

driesdeproost: I will add it to the method name, together with the `throws exception`

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
```diff
@@ -322,4 +321,31 @@ private void withLocale(Locale locale, AuthenticationUtil.RunAsWork<Object> work
      */
     abstract protected void withTestFacets(AuthenticationUtil.RunAsWork<Object> work) throws Exception;
 
+
+    @Test
+    public void testNotFilteredOut() {
+        SearchSyntaxNode node =
+                new QueryBuilder()
+                        .startOr()
+                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", "9223372036854775807")
```
kpiot123, line 330: Can you use Long.MAX_VALUE instead of "9223372036854775807" so people know what the value means?

apix-impl/src/test/java/eu/xenit/apix/tests/search/FtsNodeVisitorTest.java
```diff
@@ -0,0 +1,96 @@
...
+        PropertyService propertyService = mock(PropertyService.class);
```
kpiot123, line 77: @driesdeproost PropertyService is an service of Alfred API so you don't need to mock it. It is better to create a stub like Toon said and then mock the Alfresco services you need. In this case you will need a mock for the dictionary service I believe.

CHANGELOG.md
driesdeproost: done

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/PropertyTypeCheckService.java
```diff
@@ -0,0 +1,69 @@
...
+    public boolean fitsType(PropertySearchNode node) {
+        if (node == null || node.getValue() == null || node.getName() == null || propertyService == null) {
+            return true;
+        }
+        QName qName = new QName(node.getName().replaceAll("\\\\", "")); //stuff like hyphen in node-dbid is escaped
+        PropertyDefinition propertyDefinition = propertyService.GetPropertyDefinition(qName);
+        if (propertyDefinition == null || propertyDefinition.getDataType() == null) {
+            return true;
+        }
+        for (Entry<String, Predicate<String>> entry : dataTypePredicates.entrySet()) {
```
RVanhuysseXenit, line 43: Why do you iterate over the entrySet and compare keyvalues, instead of just calling `dataTypePredicates.get()`?

driesdeproost: That's better :+1:

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/PropertyTypeCheckService.java
driesdeproost: Didn't find anything with some quick searches
I abstracted out the existing escape mechanism and added an inverse (and some tests on them) as a proposal
https://github.com/xenit-eu/alfred-api/pull/44/commits/2aac5d36004c0c3b1c02f51e9a571f451931d321

CHANGELOG.md
tgeens: Imo we should drop the "in FtsNodeVisitor", but maybe @kpiot123 or @WimCrols can give their opionion

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/PropertyTypeCheckService.java
tgeens: See comment on that commit.

CHANGELOG.md
kpiot123: Yes, you can drop that part. People reading this don't know what an FtsNodeVisitor is.

apix-impl/src/test/java/eu/xenit/apix/tests/search/FtsNodeVisitorTest.java
```diff
@@ -0,0 +1,108 @@
+package eu.xenit.apix.tests.search;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.mockito.Mockito.mock;
```
kpiot123, line 6: @driesdeproost You don't seem to be using mocks anymore so you can remove this import

CHANGELOG.md
driesdeproost: Actually, what you see here is the outdated
Have a look at https://github.com/xenit-eu/alfred-api/pull/44/files

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/PropertyTypeCheckService.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Map;
+import java.util.function.Predicate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PropertyTypeCheckService {
+
+    private final static Logger logger = LoggerFactory.getLogger(PropertyTypeCheckService.class);
+
+    private final Map<String, Predicate<String>> dataTypePredicates = new HashMap<String, Predicate<String>>() {{
```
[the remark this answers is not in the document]
driesdeproost: Thanks, nice catch

PR #45, opened 2020-07-09
ALFREDAPI-412 handle null and not authorized cases in getAllNodeInfo

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-412

-> getAllNodeInfo may now return 400 responses instead of unwarranted 500 or faulty 200 responses.

### DIFF
--- a/apix-impl/src/integration-test/java/eu/xenit/apix/rest/v2/tests/AllNodeInfoTest.java
+++ b/apix-impl/src/integration-test/java/eu/xenit/apix/rest/v2/tests/AllNodeInfoTest.java
@@ -79,6 +79,45 @@ public class AllNodeInfoTest extends eu.xenit.apix.rest.v2.tests.BaseTest {
         }
     }
 
+    @Test
+    public void testGetAllNodeInfoWithNoNodesListed() throws IOException {
+        String jsonString = json("{}");
+
+        final CloseableHttpClient httpclient = HttpClients.createDefault();
+        final String url = makeAlfrescoBaseurl("admin", "admin") + "/apix/v2/nodes/nodeInfo";
+        logger.error("url: " + url);
+        HttpPost httppost = new HttpPost(url);
+        httppost.setEntity(new StringEntity(jsonString));
+
+        try (CloseableHttpResponse response = httpclient.execute(httppost)) {
+            assertEquals(400, response.getStatusLine().getStatusCode());
+        }
+    }
+
+    @Test
+    public void testGetAllNodeInfoForNodeWithoutPermissions() throws IOException {
+        HashMap<String, NodeRef> initializedNodeRefs = init();
+        String url =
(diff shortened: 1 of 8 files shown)


apix-rest-v1/src/main/java/eu/xenit/apix/rest/v2/ApixV2Webscript.java
```diff
@@ -44,6 +44,17 @@
             boolean retrieveTargetAssocs, boolean retrieveSourceAssocs) {
         List<NodeInfo> nodeInfoList = new ArrayList<NodeInfo>();
         for (NodeRef nodeRef : nodeRefs) {
+
+            if (!permissionService.hasPermission(nodeRef, IPermissionService.READ)) {
+                logger.warn("Excluding node {} from results due to insufficient permissions", nodeRef);
+                continue;
+            }
+
+            if (!nodeService.exists(nodeRef)) {
```
Ranecdev, line 53: Maybe change the order of these checks. NodeService.exists() is the shorter operation

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/ApixV1Webscript.java
```diff
@@ -104,6 +104,11 @@ protected NodeInfo nodeRefToNodeInfo(NodeRef nodeRef,
             return null;
         }
 
+        if (!nodeService.exists(nodeRef)) {
```
Ranecdev, line 107: See comment below

apix-impl/src/integration-test/java/eu/xenit/apix/rest/v2/tests/AllNodeInfoTest.java
```diff
@@ -79,6 +80,49 @@ public void testGetAllNodeInfoOfMultipleNodes() throws IOException, JSONExceptio
...
+        final String url = makeAlfrescoBaseurl("admin", "admin") + "/apix/v2/nodes/nodeInfo";
+        logger.error("url: " + url);
+        HttpPost httppost = new HttpPost(url);
+        httppost.setEntity(new StringEntity(jsonString));
+
+        try (CloseableHttpResponse response = httpclient.execute(httppost)) {
+            assertEquals(400, response.getStatusLine().getStatusCode());
+        }
+    }
+
```
Ranecdev, line 97: Maybe add a test for some non-existing and no-permissions nodes together with some real nodes and assert the amount of nodeInfo's returned

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v2/ApixV2Webscript.java
RVanhuysseXenit: In effect nothing changes by changing the order in this particular case, but the exists check will apparently always return true if a user has insufficient permissions to check for that node (dixit axel faust). Given this I'd keep it as is.

PR #60, opened 2020-08-05
ALFREDAPI-338 Changed CopyFrom behaviour to also set the provide properties

Ranecdev:
…erties in the request

Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-338

### DIFF
--- a/apix-impl/src/integration-test/java/eu/xenit/apix/rest/v1/tests/CopyNodeTest.java
+++ b/apix-impl/src/integration-test/java/eu/xenit/apix/rest/v1/tests/CopyNodeTest.java
@@ -57,7 +57,7 @@ public class CopyNodeTest extends BaseTest {
 
         transactionService.getRetryingTransactionHelper()
                 .doInTransaction(() -> {
-                    doTestCopy(httpclient, url, parentRef, initializedNodeRefs.get(BaseTest.TESTFILE_NAME).toString(), 200);
+                    doTestCopy(httpclient, url, parentRef, initializedNodeRefs.get(BaseTest.TESTFILE_NAME).toString(),200);
                     return null;
                 }, false, true);
 
--- a/apix-impl/src/integration-test/java/eu/xenit/apix/rest/v1/tests/CopyNodeTest.java
+++ b/apix-impl/src/integration-test/java/eu/xenit/apix/rest/v1/tests/CopyNodeTest.java
@@ -1,19 +1,25 @@
 package eu.xenit.apix.rest.v1.tests;
 
 import eu.xenit.apix.data.NodeRef;
+import eu.xenit.apix.data.QName;
 import eu.xenit.apix.node.INodeService;
 import eu.xenit.apix.node.NodeAssociation;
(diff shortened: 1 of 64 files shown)


apix-impl/src/integration-test/java/eu/xenit/apix/rest/v1/tests/CopyNodeTest.java
```diff
@@ -77,30 +107,69 @@ public void copyNodeReturnsAccesDenied() {
...
+                    doTestCopy(httpclient, url, parentRef, initializedNodeRefs.get(BaseTest.NOUSERRIGHTS_FILE_NAME).toString(),null, null, 403);
                     return null;
                 }, false, true);
 
         List<ChildParentAssociation> newChildAssocs = nodeService.getChildAssociations(parentRef);
         assertEquals(1, newChildAssocs.size());
     }
 
-    private void doTestCopy(CloseableHttpClient httpClient, String url, NodeRef parentRef, String targetRef, int expectedResponseCode) throws Throwable {
+    private String doTestCopy(CloseableHttpClient httpClient, String url, NodeRef parentRef, String copyFrom, String name, HashMap<QName, String[]> properties, int expectedResponseCode) throws Throwable {
```
todorinskiz, line 118: Please use POJOs that are later serialized with e.g. Newtonsoft ...

apix-impl/src/integration-test/java/eu/xenit/apix/rest/v1/tests/CopyNodeTest.java
```diff
@@ -52,19 +61,49 @@ public void testCopyNode() {
         final ChildParentAssociation primaryParentAssoc = (ChildParentAssociation) parentAssociations.get(0);
         final NodeRef parentRef = primaryParentAssoc.getTarget();
 
-        final String url = makeAlfrescoBaseurl("admin", "admin") + "/apix/v1/nodes";
+        final String url = makeAlfrescoBaseurl("admin", "admin") + "/apix/v2/nodes";
```
koenvangeerteruy, line 64: @Ranecdev The test now points to v2 of the REST API but these are a test set for the v1 REST API.
Changes have been made to v1 and v2 webscripts. Test should be provided for both v1 and v2 REST endpoints
Tests on v1 should also contain requests with the previous set of parameters since we also want to test we didn't brake the API

Ranecdev: > Tests on v1 should also contain requests with the previous set of parameters since we also want to test we didn't brake the API

Should this new behavior only be added to v2? (Also their are some olders tests that are still functioning)

koenvangeerteruy: I believe this change is a fix, so that the endpoint works as intended. I think it should be fixed in both versions.
I just want to make sure that a previous request format to the v1 endpoint still keeps working

PR #68, opened 2020-08-12
Alfredapi 298 - Add CRUD api for comments

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-298

This change adds a new interface to the java api and extends the rest api with 3 new endpoints

### DIFF
--- a/apix-impl/50/src/main/java/eu/xenit/apix/alfresco50/CommentServiceImpl50.java
+++ b/apix-impl/50/src/main/java/eu/xenit/apix/alfresco50/CommentServiceImpl50.java
@@ -1,0 +1,94 @@
+package eu.xenit.apix.alfresco50;
+
+import com.github.dynamicextensionsalfresco.osgi.OsgiService;
+import eu.xenit.apix.alfresco.ApixToAlfrescoConversion;
+import eu.xenit.apix.alfresco.comments.CommentService;
+import eu.xenit.apix.comments.Comment;
+import eu.xenit.apix.content.IContentService;
+import eu.xenit.apix.node.INodeService;
+import eu.xenit.apix.permissions.IPermissionService;
+import java.io.Serializable;
+import java.util.Set;
+import org.alfresco.model.ContentModel;
+import org.alfresco.repo.security.authentication.AuthenticationUtil;
+import org.alfresco.repo.site.SiteModel;
+import org.alfresco.service.cmr.lock.LockService;
+import org.alfresco.service.cmr.lock.LockStatus;
+import org.alfresco.service.cmr.repository.NodeRef;
+import org.alfresco.service.cmr.repository.NodeService;
+import org.alfresco.service.cmr.security.AccessStatus;
+import org.alfresco.service.cmr.security.PermissionService;
+import org.alfresco.service.namespace.QName;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service("eu.xenit.apix.comments.CommentService")
+@OsgiService
+public class CommentServiceImpl50 extends CommentService {
+
+    private static final Logger log = LoggerFactory.getLogger(CommentServiceImpl50.class);
+
+    protected LockService alfLockService;
+    protected NodeService alfNodeService;
+    protected PermissionService alfPermissionService;
+
+    @Autowired
+    public CommentServiceImpl50(
+            LockService alfLockService,
+            NodeService alfNodeService,
+            PermissionService alfPermissionService,
+            org.alfresco.repo.forum.CommentService commentService,
+            IContentService contentService, INodeService nodeService,
+            IPermissionService permissionService,
+            ApixToAlfrescoConversion apixConverter) {
+        super(commentService, contentService, nodeService, permissionService, apixConverter);
+        this.alfLockService = alfLockService;
+        this.alfNodeService = alfNodeService;
+        this.alfPermissionService = alfPermissionService;
+    }
+
+    @Override
+    protected Comment setPermissions(NodeRef documentNode,
+            NodeRef commentNodeRef, Comment targetComment) {
+        boolean canEdit = false;
+        boolean canDelete = false;
+        if(!isWorkingCopyOrLocked(documentNode)) {
+            canEdit = canEditPermission(commentNodeRef);
+            canDelete = canDeletePermission(commentNodeRef);
+        }
+        targetComment.setCanEdit(canEdit);
+        targetComment.setCanDelete(canDelete);
+        return targetComment;
+    }
+
+    protected boolean isWorkingCopyOrLocked(NodeRef documentNode) {
+        boolean isworkingcopy = false;
+        boolean islocked = false;
+        if (alfNodeService.exists(documentNode)) {
+            Set<QName> documentAspects = alfNodeService.getAspects(documentNode);
+            isworkingcopy = documentAspects.contains(ContentModel.ASPECT_WORKING_COPY);
+            if (!isworkingcopy && documentAspects.contains(ContentModel.ASPECT_LOCKABLE)) {
+                log.debug("Node is not a working copy");
+                LockStatus lockStatus = alfLockService.getLockStatus(documentNode);
+                islocked = lockStatus == LockStatus.LOCKED;
+            }
+        }
+        return (isworkingcopy || islocked);
+    }
+
+    protected boolean canEditPermission(NodeRef commentNodeRef) {
+        String creator = (String) alfNodeService.getProperty(commentNodeRef, ContentModel.PROP_CREATOR);
+        Serializable owner = alfNodeService.getProperty(commentNodeRef, ContentModel.PROP_OWNER);
+        String currentUser = AuthenticationUtil.getFullyAuthenticatedUser();
+        boolean isSiteManager = alfPermissionService.hasPermission(commentNodeRef, SiteModel.SITE_MANAGER) == (AccessStatus.ALLOWED);
+        boolean isCoordinator = alfPermissionService.hasPermission(commentNodeRef, PermissionService.COORDINATOR) == (AccessStatus.ALLOWED);
+        return (isSiteManager || isCoordinator || currentUser.equals(creator) || currentUser.equals(owner));
+    }
+
+    protected boolean canDeletePermission(NodeRef commentNodeRef)
+    {
+        return alfPermissionService.hasPermission(commentNodeRef, PermissionService.DELETE) == AccessStatus.ALLOWED;
+    }
+}
--- a/apix-impl/50/src/main/java/eu/xenit/apix/alfresco50/CommentServiceImpl50.java
+++ b/apix-impl/50/src/main/java/eu/xenit/apix/alfresco50/CommentServiceImpl50.java
@@ -54,12 +54,12 @@ public class CommentServiceImpl50 extends CommentService {
             NodeRef commentNodeRef, Comment targetComment) {
         boolean canEdit = false;
         boolean canDelete = false;
-        if(!isWorkingCopyOrLocked(documentNode)) {
+        if (!isWorkingCopyOrLocked(documentNode)) {
             canEdit = canEditPermission(commentNodeRef);
             canDelete = canDeletePermission(commentNodeRef);
         }
-        targetComment.setCanEdit(canEdit);
-        targetComment.setCanDelete(canDelete);
+        targetComment.setEditable(canEdit);
+        targetComment.setDeletable(canDelete);
         return targetComment;
     }
 
@@ -82,13 +82,14 @@ public class CommentServiceImpl50 extends CommentService {
         String creator = (String) alfNodeService.getProperty(commentNodeRef, ContentModel.PROP_CREATOR);
         Serializable owner = alfNodeService.getProperty(commentNodeRef, ContentModel.PROP_OWNER);
         String currentUser = AuthenticationUtil.getFullyAuthenticatedUser();
-        boolean isSiteManager = alfPermissionService.hasPermission(commentNodeRef, SiteModel.SITE_MANAGER) == (AccessStatus.ALLOWED);
-        boolean isCoordinator = alfPermissionService.hasPermission(commentNodeRef, PermissionService.COORDINATOR) == (AccessStatus.ALLOWED);
+        boolean isSiteManager =
+                alfPermissionService.hasPermission(commentNodeRef, SiteModel.SITE_MANAGER) == (AccessStatus.ALLOWED);
+        boolean isCoordinator = alfPermissionService.hasPermission(commentNodeRef, PermissionService.COORDINATOR)
+                == (AccessStatus.ALLOWED);
         return (isSiteManager || isCoordinator || currentUser.equals(creator) || currentUser.equals(owner));
     }
 
-    protected boolean canDeletePermission(NodeRef commentNodeRef)
-    {
+    protected boolean canDeletePermission(NodeRef commentNodeRef) {
         return alfPermissionService.hasPermission(commentNodeRef, PermissionService.DELETE) == AccessStatus.ALLOWED;
     }
 }
--- a/apix-impl/51/src/main/java/eu/xenit/apix/alfresco51/CommentServiceImpl51.java
+++ b/apix-impl/51/src/main/java/eu/xenit/apix/alfresco51/CommentServiceImpl51.java
@@ -1,0 +1,94 @@
+package eu.xenit.apix.alfresco51;
+
+import com.github.dynamicextensionsalfresco.osgi.OsgiService;
+import eu.xenit.apix.alfresco.ApixToAlfrescoConversion;
+import eu.xenit.apix.alfresco.comments.CommentService;
+import eu.xenit.apix.comments.Comment;
+import eu.xenit.apix.content.IContentService;
+import eu.xenit.apix.node.INodeService;
+import eu.xenit.apix.permissions.IPermissionService;
+import java.io.Serializable;
+import java.util.Set;
+import org.alfresco.model.ContentModel;
+import org.alfresco.repo.security.authentication.AuthenticationUtil;
+import org.alfresco.repo.site.SiteModel;
+import org.alfresco.service.cmr.lock.LockService;
+import org.alfresco.service.cmr.lock.LockStatus;
+import org.alfresco.service.cmr.repository.NodeRef;
+import org.alfresco.service.cmr.repository.NodeService;
+import org.alfresco.service.cmr.security.AccessStatus;
+import org.alfresco.service.cmr.security.PermissionService;
+import org.alfresco.service.namespace.QName;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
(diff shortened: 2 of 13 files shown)


apix-impl/src/main/java/eu/xenit/apix/alfresco/comments/CommentService.java
```diff
@@ -0,0 +1,134 @@
...
+    protected Comment toComment(org.alfresco.service.cmr.repository.NodeRef documentNode,
+            org.alfresco.service.cmr.repository.NodeRef commentNodeRef) {
+        NodeRef apixCommentNodeRef = apixConverter.apix(commentNodeRef);
+        NodeMetadata commentMetadata = nodeService.getMetadata(apixCommentNodeRef);
+        String content;
+        try {
+            content = IOUtils.toString(contentService.getContent(apixCommentNodeRef).getInputStream());
+        } catch (IOException ioException) {
+            log.error("Error reading content for comment {}", commentNodeRef, ioException);
+            content = "Comment not available: error reading content";
```
vierbergenlars, line 101: Just throw the exception upwards when we can't read comments.
It's a bad idea to replace the comment contents with an error message, because then you can't distinguish between a comment and a comment that failed to be read.

apix-impl/51/src/main/java/eu/xenit/apix/alfresco51/CommentServiceImpl51.java
```diff
@@ -0,0 +1,94 @@
...
+@OsgiService
+public class CommentServiceImpl51 extends CommentService {
+
+    private static final Logger log = LoggerFactory.getLogger(CommentServiceImpl51.class);
+
+    protected ServiceRegistry serviceRegistry;
+
+    @Autowired
+    public CommentServiceImpl51(
+            ServiceRegistry serviceRegistry,
```
vierbergenlars, line 38: Instead of injecting `ServiceRegistry`, we usually inject the necessary service itself

apix-interface/src/main/java/eu/xenit/apix/comments/Conversation.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public List<Comment> getComments() {
+        return comments;
+    }
+
+    public void setComments(List<Comment> comments) {
+        this.comments = comments;
+    }
+
+    public Conversation comments(List<Comment> comments) {
```
vierbergenlars, line 20: If you want to use a builder pattern, I would have expected a more explicit implementation of that, with a separate `ConversationBuilder` class.

RVanhuysseXenit: This is more out of a desire for something like properties rather than an actual builder (that seems overkill). Ill remove them to keep the code clean

apix-impl/src/main/java/eu/xenit/apix/alfresco/comments/CommentService.java
```diff
@@ -96,9 +96,10 @@ protected Comment toComment(org.alfresco.service.cmr.repository.NodeRef document
         String content;
         try {
             content = IOUtils.toString(contentService.getContent(apixCommentNodeRef).getInputStream());
-        } catch (IOException ioException) {
-            log.error("Error reading content for comment {}", commentNodeRef, ioException);
-            content = "Comment not available: error reading content";
+        } catch (IOException e) {
+            String message = String.format("Encountered an IOException while handling comment %s", commentNodeRef);
+            log.error(message, e);
+            throw new AlfrescoRuntimeException(message, e);
```
vierbergenlars, line 102: If we throw an exception, there is no need to log it. It will also be logged when it's not caught higher up, leading to 2 loglines for the same error.

apix-impl/52/src/main/java/eu/xenit/apix/alfresco52/CommentServiceImpl52.java
```diff
@@ -0,0 +1,34 @@
...
+    public CommentServiceImpl52(org.alfresco.repo.forum.CommentService commentService,
+            IContentService contentService, INodeService nodeService,
+            IPermissionService permissionService,
+            ApixToAlfrescoConversion apixConverter) {
+        super(commentService, contentService, nodeService, permissionService, apixConverter);
+    }
+
+    @Override
+    protected Comment setPermissions(org.alfresco.service.cmr.repository.NodeRef documentNode,
+            org.alfresco.service.cmr.repository.NodeRef commentNodeRef, Comment targetComment) {
```
kpiot123, line 28: Is there a reason why in `CommentServiceImpl50` and `CommentServiceImpl51` `org.alfresco.service.cmr.repository.NodeRef` is imported but in `CommentServiceImpl52`, `CommentServiceImpl60` and `CommentServiceImpl61` ir isn't?

apix-impl/50/src/main/java/eu/xenit/apix/alfresco50/CommentServiceImpl50.java
```diff
@@ -0,0 +1,94 @@
...
+        String creator = (String) alfNodeService.getProperty(commentNodeRef, ContentModel.PROP_CREATOR);
+        Serializable owner = alfNodeService.getProperty(commentNodeRef, ContentModel.PROP_OWNER);
+        String currentUser = AuthenticationUtil.getFullyAuthenticatedUser();
+        boolean isSiteManager = alfPermissionService.hasPermission(commentNodeRef, SiteModel.SITE_MANAGER) == (AccessStatus.ALLOWED);
+        boolean isCoordinator = alfPermissionService.hasPermission(commentNodeRef, PermissionService.COORDINATOR) == (AccessStatus.ALLOWED);
+        return (isSiteManager || isCoordinator || currentUser.equals(creator) || currentUser.equals(owner));
+    }
+
+    protected boolean canDeletePermission(NodeRef commentNodeRef)
+    {
```
rschev, line 91: Shouldn't be on a separate line. Same for 51.

apix-impl/src/main/java/eu/xenit/apix/alfresco/comments/CommentService.java
```diff
@@ -0,0 +1,134 @@
...
+import org.alfresco.query.PagingRequest;
+import org.alfresco.query.PagingResults;
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class CommentService implements ICommentService {
+
+    private static final Logger log = LoggerFactory.getLogger(CommentService.class);
```
rschev, line 27: Logger isn't used.

apix-impl/src/main/java/eu/xenit/apix/alfresco/comments/CommentService.java
```diff
@@ -0,0 +1,134 @@
...
+        commentService.deleteComment(apixConverter.alfresco(targetCommentNode));
+    }
+
+    @Override
+    public boolean canCreateComment(NodeRef targetNodeRef) {
+        return permissionService.hasPermission(targetNodeRef, IPermissionService.CREATE_CHILDREN);
+    }
+
+    protected Comment toComment(org.alfresco.service.cmr.repository.NodeRef documentNode,
+            org.alfresco.service.cmr.repository.NodeRef commentNodeRef) {
```
rschev, line 93: I would expect most of this method to be part of a constructor for `Comment`. The alf→apix conversion and IOUtils stuff is probably out of scope, but the rest seems like it's behaviour that deals with the state of `Comment`, not the state of `CommentService`, so it should go in `Comment`.

apix-interface/src/main/java/eu/xenit/apix/comments/Comment.java
```diff
@@ -0,0 +1,109 @@
...
+        this.canEdit = canEdit;
+    }
+
+    public boolean isCanDelete() {
+        return canDelete;
+    }
+
+    public void setCanDelete(boolean canDelete) {
+        this.canDelete = canDelete;
+    }
```
rschev, line 88: The names of these methods are kind of strange. I know that `isSomething` is a convention for boolean getters, but `isCanDelete` and `setCanDelete` sound like you really want to say `isDeletable`/`setDeletable`.

apix-interface/src/main/java/eu/xenit/apix/comments/Conversation.java
```diff
@@ -0,0 +1,64 @@
...
+        this.hasMore = hasMore;
+    }
+
+    public boolean isCanCreate() {
+        return canCreate;
+    }
+
+    public void setCanCreate(boolean canCreate) {
+        this.canCreate = canCreate;
+    }
```
rschev, line 43: Same note about the naming here.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -651,6 +657,96 @@ public void setParent(@UriVariable final String space, @UriVariable final String
...
+            @ApiResponse(code = 200, message = "Success"),
+            @ApiResponse(code = 403, message = "Not Authorized"),
+            @ApiResponse(code = 404, message = "Not Found")
+    })
+    public void postComment(@UriVariable String space, @UriVariable String store, @UriVariable String guid,
+            WebScriptRequest request, WebScriptResponse response) throws IOException {
+        final NodeRef target = new NodeRef(space, store, guid);
+        if (nodeService.exists(target)) {
+            ObjectMapper objectMapper = new ObjectMapper();
+            String content = objectMapper.readTree(request.getContent().getInputStream()).textValue();
```
rschev, line 696: I'm a bit confused about the purpose of this line. We're reading in the content of the request as a JsonNode, just to get the text value of it. Isn't that equivalent to stripping off the leading and trailing quotation marks of the content?

What json structure is the payload of this endpoint supposed to conform to anyway? Is there a swagger spec for this?

apix-impl/src/main/java/eu/xenit/apix/alfresco/comments/CommentService.java
```diff
@@ -0,0 +1,134 @@
...
+import org.alfresco.error.AlfrescoRuntimeException;
+import org.alfresco.model.ContentModel;
+import org.alfresco.query.PagingRequest;
+import org.alfresco.query.PagingResults;
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class CommentService implements ICommentService {
```
rschev, line 25: Shouldn't this be abstract? Every version extends it.

apix-impl/src/main/java/eu/xenit/apix/alfresco/comments/CommentService.java
```diff
@@ -0,0 +1,134 @@
...
+            response.setModifiedBy(property.get(0));
+        }
+        setPermissions(documentNode, commentNodeRef, response);
+        return response;
+    }
+
+    protected Comment setPermissions(org.alfresco.service.cmr.repository.NodeRef documentNode,
+            org.alfresco.service.cmr.repository.NodeRef commentNodeRef, Comment targetComment) {
+        return targetComment;
+    }
```
rschev, line 133: This method doesn't do anything and is overrided everywhere, I think it wants to be an abstract method instead.

apix-interface/src/main/java/eu/xenit/apix/comments/Comment.java
RVanhuysseXenit: they were autogenerated, but I suppose I can clean it up

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
RVanhuysseXenit: there is a spec on the ticket. Im not entirely sure where to put it in the repo though

RVanhuysseXenit: I dont get the remark over stripping though?

vierbergenlars: > Isn't that equivalent to stripping off the leading and trailing quotation marks of the content?

No, because if you have `"some text\"some more text"`, this must decode to `some text"some more text`, and not to `some text\"some more text`.

apix-interface/src/main/java/eu/xenit/apix/comments/Comment.java
vierbergenlars: I'm not sure how jackson would like those other names when serializing to json.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
vierbergenlars: Also, this is not according to spec. 
The request  body is JSON and should have a `content` property (a very trimmed down version of the  `Comment` class that only has the `content` field.)

rschev: After reading the specand what Lars said it makes more sense. 
It's probably not worth another POJO but it should be something like 
```
JsonNode jsonNode = objectMapper.readTree(request.getContent().getInputStream());
if (jsonNode.has("content")) {
    String content = jsonNode.get("content").textValue();
    // ... continue
} else {
    // send an error about the missing field
}
```

vierbergenlars: Not using a POJO will break the automatic generation of a proper swagger spec from apix. (Or are we no longer doing that?)

apix-impl/src/main/java/eu/xenit/apix/alfresco/comments/CommentService.java
RVanhuysseXenit: im not a fan of putting the disentanglement of the properties from the property map in the pojo class, since the properties are extracted based on QName which is strictly alfresco, while the apix interface should not be dependant on alfresco. The alternative would be a builder, but seeing as I wouldnt want that in the interface either, it would end up in the impl module again.

apix-interface/src/main/java/eu/xenit/apix/comments/Comment.java
RVanhuysseXenit: I can refactor the property names to `deletable` . that should clean up somewhat and still work.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
rschev: I'm not sure, but it's probably good to keep that option open, so using a POJO instead of what I wrote.

apix-impl/src/main/java/eu/xenit/apix/alfresco/comments/CommentService.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+        property = commentMetadata.properties.get(new QName(ContentModel.PROP_CREATOR.toString()));
+        if (property != null && !property.isEmpty()) {
+            response.setCreatedBy(property.get(0));
+        }
+        property = commentMetadata.properties.get(new QName(ContentModel.PROP_MODIFIED.toString()));
+        if (property != null && !property.isEmpty()) {
+            response.setModifiedAt(property.get(0));
+        }
+        property = commentMetadata.properties.get(new QName(ContentModel.PROP_MODIFIER.toString()));
```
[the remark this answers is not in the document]
RVanhuysseXenit: Changing that would best be done by adding a new method on the nodeservice to return a `map<string, serialiazble>`. we might also then just filrter out the lists for non-multivalue properties as well. but I consider the nodeservice to be out of scope for this ticket.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -651,6 +657,88 @@ public void setParent(@UriVariable final String space, @UriVariable final String
...
+    }
+
+    @ApiOperation(value = "Appends a new comment to the given node.")
+    @Uri(value = "/nodes/{space}/{store}/{guid}/comments", method = HttpMethod.POST)
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "Success"),
+            @ApiResponse(code = 403, message = "Not Authorized"),
+            @ApiResponse(code = 404, message = "Not Found")
+    })
+    public void postComment(@UriVariable String space, @UriVariable String store, @UriVariable String guid,
```
WimCrols, line 691: (non-blocking) If the other method is called `updateComment`, this one should be called `createComment` or `addComment` IMO.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -651,6 +657,88 @@ public void setParent(@UriVariable final String space, @UriVariable final String
...
+        } else {
+            writeNotFoundResponse(response, targetComment);
+        }
+    }
+
+    @ApiOperation(value = "Deletes a given comment on the given node.")
+    @Uri(value = "/nodes/{space}/{store}/{guid}/comment",
+            method = HttpMethod.DELETE)
+    @ApiResponses({
+            @ApiResponse(code = 204, message = "Success"),
```
WimCrols, line 726: **204** should not be used. Instead use one of https://github.com/xenit-eu/alfred-api/blob/master/developer-documentation/developer.md#2xx-success

WimCrols: Idem line 735

apix-interface/src/main/java/eu/xenit/apix/comments/Conversation.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public List<Comment> getComments() {
+        return comments;
+    }
+
+    public void setComments(List<Comment> comments) {
+        this.comments = comments;
+    }
+
+    public boolean isHasMore() {
```
WimCrols, line 29: isHasMore -> hasMore

rschev (approved):

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
RVanhuysseXenit: 200 then.

apix-interface/src/main/java/eu/xenit/apix/comments/Conversation.java
RVanhuysseXenit: Lars made a note earlier that changing the format might mess with serialization/deserialization? Still requires change, or keep as is?

rschev (approved):

apix-interface/src/main/java/eu/xenit/apix/comments/Conversation.java
WimCrols: No, you can leave it then if has impact.

WimCrols (approved):

PR #73, opened 2020-08-26
Alfredapi 403 Added support for special search terms 'isunset', 'isnull', 'isnotnull' and 'exists'

kpiot123:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-403


apix-impl/src/test/java/eu/xenit/apix/tests/search/FtsNodeVisitorTest.java
```diff
@@ -57,6 +63,42 @@ public void testIntTypeValid() {
                 containsString("{tenant.model}intProperty"));
     }
 
+    @Test
+    public void testIsUnsetTerm() {
+        String ftsQuery = convertToFtsTerm(IS_UNSET);
```
WimCrols, line 68: Nice short tests :+1: 
You could even write `assertEquals("ISUNSET:\"prefix:test_prop\"", convertToFtsTerm(IS_UNSET));`, but that's a matter of preference and you don't have to change it.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/search/SearchWebScript1.java
```diff
@@ -47,6 +47,19 @@
...
           + "Info about the Search query syntax can be found here: https://xenitsupport.jira.com/wiki/display/APIX/Search+query+syntax\n"
+          + "### special search terms:\n"
+          + "- type: searches for nodes of that type (for example: \"type\" : \"cm:content\")\n"
+          + "- aspect: searches for nodes with that aspect (for example: \"aspect\" : \"cm:titled\")\n"
+          + "- noderef: searches the node with that noderef (for example: \"noderef\" : \"workspace://SpacesStore/f0d15919-3841-4170-807f-b81d2ebdeb80\")\n"
+          + "- parent: searches the nodes with that parent (for example: \"parent\" : \"workspace://SpacesStore/f0d15919-3841-4170-807f-b81d2ebdeb80\")\n"
+          + "- path: searches the nodes with that path (for example: \"path\" : \"/\")\n"
+          + "- category: searches the nodes with that category (for example: \"category\" : \"workspace://SpacesStore/f0d15919-3841-4170-807f-b81d2ebdeb80\")\n"
+          + "- text: searches the nodes with content containing that text (for example: \"text\" : \"this text\")\n"
+          + "- all: searches the nodes with content containing the value (for example: \"all\" : \"search term\")\n"
```
koenvangeerteruy, line 58: @kpiot123 Documentation for all is not complete. Not only is searched in content but also in name and author
https://github.com/xenit-eu/alfred-api/blob/8699ed6ef671a4c9766c289a1632fe7c91e1ac4f/apix-impl/src/main/java/eu/xenit/apix/alfresco/search/FtsNodeVisitor.java#L110

apix-impl/src/main/java/eu/xenit/apix/alfresco/search/LuceneNodeVisitor.java
```diff
@@ -26,6 +26,10 @@ public LuceneNodeVisitor() {
         termDictionary.put("category", "CATEGORY");
         termDictionary.put("text", "TEXT");
         termDictionary.put("all", "ALL");
+        termDictionary.put("isunset", "ISUNSET");
```
WimCrols, line 29: I was trying to understand why there are **two** dictionaries like this, so I searched where _LuceneNodeVisitor_ is used. I can't find any usages!
Why did you add it here? Maybe I'm missing something?

WimCrols (changes requested):

apix-impl/src/main/java/eu/xenit/apix/alfresco/search/LuceneNodeVisitor.java
koenvangeerteruy: I think is was added for completeness.
When Lucene is not supported anymore by the search backend in Alfresco 5.x - 6.x then it can be removed.
That clean up should have been part of the removal of support for Alfresco 4.2. Unless impact is low the LuceneNodeVisitor can be removed in this ticket otherwise I would suggest to create a ticket to clean up unused code

kpiot123: hmm, I added it because I was looking where the other keywords were used and added the new ones to it. Since LuceneNodeVisitor is only used in tests I don't have to add the new keywords there. I will revert these changes.

I think the LuceneNodeVisitor was added when we were still supporting Alfresco 4.1 and 4.2 but since we don't do that anymore we also don't need that visitor anymore.

kpiot123: I will create a ticket for the removal of the LuceneNodeVisitor and its related tests.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/search/SearchWebScript1.java
kpiot123: I must have been distracted while I was writing that

WimCrols (dismissed):

koenvangeerteruy (approved):

PR #74, opened 2020-08-26
ALFREDAPI-442 implement webscript for /properties and /aspects

rawiabenhmida:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-442

### DIFF
--- a/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
@@ -4,8 +4,10 @@ import com.github.dynamicextensionsalfresco.osgi.OsgiService;
 import eu.xenit.apix.alfresco.ApixToAlfrescoConversion;
 import eu.xenit.apix.data.QName;
 import eu.xenit.apix.dictionary.aspects.AspectDefinition;
+import eu.xenit.apix.dictionary.aspects.Aspects;
 import eu.xenit.apix.dictionary.aspects.IAspectService;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import org.alfresco.service.cmr.dictionary.DictionaryService;
 import org.slf4j.Logger;
@@ -54,6 +56,20 @@ public class AspectService implements IAspectService {
             return null;
         }
         return GetAspectDefinition(c.alfresco((qname)));
+    }
+
+
+    @Override
+    public Aspects getAspects() {
+        Collection<org.alfresco.service.namespace.QName> aspects = dictionaryService
+                .getAllAspects();
+        List<AspectDefinition> ret = new ArrayList<>(aspects.size());
+        for (org.alfresco.service.namespace.QName aspect : aspects) {
+            AspectDefinition aspectDefinition = GetAspectDefinition(aspect);
+            assert aspectDefinition != null;
+            ret.add(aspectDefinition);
+        }
+        return new Aspects(ret);
     }
 
     private boolean HasAspectDefinition(org.alfresco.service.namespace.QName qname) {
--- a/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/DictionaryService.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/DictionaryService.java
@@ -5,6 +5,7 @@ import com.github.dynamicextensionsalfresco.osgi.OsgiService;
 import eu.xenit.apix.alfresco.ApixToAlfrescoConversion;
 import eu.xenit.apix.dictionary.IDictionaryService;
 import eu.xenit.apix.dictionary.aspects.AspectDefinition;
+import eu.xenit.apix.dictionary.aspects.Aspects;
 import eu.xenit.apix.dictionary.aspects.IAspectService;
 import eu.xenit.apix.dictionary.namespaces.Namespace;
 import eu.xenit.apix.dictionary.namespaces.Namespaces;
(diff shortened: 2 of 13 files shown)


apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
```diff
@@ -56,6 +58,23 @@ public AspectDefinition GetAspectDefinition(QName qname) {
...
+        for (org.alfresco.service.namespace.QName aspect : aspects) {
+
+            AspectDefinition aspectDefinition = GetAspectDefinition(aspect);
+            assert aspectDefinition != null;
+            ret.add(aspectDefinition);
+        }
+        return new Aspects(ret);
+
+
+        }
```
kpiot123, line 76: @rawiabenhmida Can you remove the new lines at the start of the for loop and underneath `return new Aspects(ret);`?
Can you also move the last "}" at the end of the method 1 tab to the left?

apix-impl/src/main/java/eu/xenit/apix/alfresco/properties/PropertyServiceImpl.java
```diff
@@ -2,12 +2,15 @@
 
 import com.github.dynamicextensionsalfresco.osgi.OsgiService;
 import eu.xenit.apix.alfresco.ApixToAlfrescoConversion;
+import eu.xenit.apix.data.QName;
```
kpiot123, line 5: @rawiabenhmida This import is not used so it can be removed.

apix-impl/src/main/java/eu/xenit/apix/alfresco/properties/PropertyServiceImpl.java
```diff
@@ -70,6 +73,45 @@ public PropertyIndexOptions GetPropertyIndexOptions(PropertyDefinition definitio
...
+        propertyDefinitionUnderConstruction.setDefaultValue(definition.getDefaultValue());
+        propertyDefinitionUnderConstruction.setDataType(c.apix(definition.getDataType().getName()));
+        propertyDefinitionUnderConstruction.setMultiValued(definition.isMultiValued());
+        propertyDefinitionUnderConstruction.setMandatory(definition.isMandatory());
+        propertyDefinitionUnderConstruction.setEnforced(definition.isMandatoryEnforced());
+        propertyDefinitionUnderConstruction.setIsProtected(definition.isProtected());
+        propertyDefinitionUnderConstruction.setIndexed(this.GetPropertyIndexOptions(definition));
+        propertyDefinitionUnderConstruction.setConstraints(this.GetConstraints(definition));
+        return propertyDefinitionUnderConstruction;
+    }
```
kpiot123, line 94: @rawiabenhmida This method looks very much like the method above it (GetPropertyDefinition). The only difference is the if statement and qname (alfresco <-> apix). So can you put the duplicate code in a separate method?

apix-interface/src/main/java/eu/xenit/apix/dictionary/aspects/Aspects.java
```diff
@@ -0,0 +1,24 @@
+package eu.xenit.apix.dictionary.aspects;
+
+import eu.xenit.apix.dictionary.types.TypeDefinition;
```
kpiot123, line 3: @rawiabenhmida This import is not used so it can be removed.

apix-interface/src/main/java/eu/xenit/apix/dictionary/properties/IPropertyService.java
```diff
@@ -1,5 +1,17 @@
 package eu.xenit.apix.dictionary.properties;
 
+import eu.xenit.apix.data.QName;
```
kpiot123, line 3: @rawiabenhmida This import is not used so it can be removed.

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/DictionaryService.java
```diff
@@ -109,9 +111,19 @@ public TypeDefinition GetTypeDefinition(eu.xenit.apix.data.QName qname) {
...
+    }
 
     @Override
     public AspectDefinition GetAspectDefinition(eu.xenit.apix.data.QName qname) {
         return aspectService.GetAspectDefinition(qname);
     }
+
+    @Override
+    public Aspects getAspects() {
+    return aspectService.getAspects();
```
WimCrols, line 126: I find it peculiar that even though we already have an AspectService, the aspect are accessed via DictionaryService, which does nothing but relay to the AspectService... However, I checked and this has been in the code since 2015, so it's not blocking here :)

apix-interface/src/main/java/eu/xenit/apix/properties/Properties.java
```diff
@@ -0,0 +1,20 @@
+package eu.xenit.apix.properties;
+
+import java.util.List;
+
+public class Properties {
```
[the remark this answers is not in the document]
vierbergenlars: You do not want an array as a top-level response in a JSON API.
It makes it impossible to extend the response with additional data (pagination/other request metadata) at a later stage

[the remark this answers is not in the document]
WimCrols: I understand. Can you then add a short comment stating that above the class?

apix-impl/src/main/java/eu/xenit/apix/alfresco/properties/PropertyServiceImpl.java
WimCrols: Thanks for the improvement, but why not put this as a constructor for PropertyDefinition in https://github.com/xenit-eu/alfred-api/blob/master/apix-interface/src/main/java/eu/xenit/apix/properties/PropertyDefinition.java ? Or are you trying to avoid a dependency on `messageService`?

WimCrols (changes requested):

apix-impl/src/main/java/eu/xenit/apix/alfresco/properties/PropertyServiceImpl.java
WimCrols: Normally, for these conversations, the ctor takes 1 argument... but we already spent a lot of time on this, so fine for now.

WimCrols (approved):

WimCrols (approved):

kpiot123 (approved):

PR #85, opened 2020-09-10
Alfredapi-418 Alfresco 6.2 support

WimCrols:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-418


apix-impl/50/src/main/java/eu/xenit/apix/alfresco50/SpringConfiguration.java
```diff
@@ -1,9 +0,0 @@
-package eu.xenit.apix.alfresco50;
```
RVanhuysseXenit, line 1: Weren't these required to make the workflow engine configurable? (I seem to remember that config was only loaded if this file was present? @rschev do you remember whether this is correct?)

apix-impl/config/alfresco-global.properties
```diff
@@ -5,3 +5,6 @@
 #aps.rest.base_url = http://process-service:8080/activiti-app/api
 #aps.rest.username = admin
 #aps.rest.password = admin
+
+#Properties to disable the org.alfresco.repo.node.ContentPropertyRestrictionInterceptor.java since it is botched
+contentPropertyRestrictions.enabled=false
```
RVanhuysseXenit, line 10: We've added this just for 6.2. Would it be reasonable to make this propertiesfile exclusive to that version of would that be to much effort. Mostly thinking of unexpected side effects in older versions?

apix-integrationtests/build.gradle
```diff
@@ -59,11 +59,9 @@ subprojects {
                 'Bundle-ClassPath': bundleClassPath(configurations.compile),
                 'Bundle-SymbolicName': "${project.group}.${project.name}",
                 'Bundle-Description': 'Alfred API: Integration tests',
-                'Import-Package': 'net.sf.cglib.core,'
```
RVanhuysseXenit, line 62: Same issue here, should we make this change specific to 6.2?

apix-impl/61/overrides.gradle
```diff
@@ -5,5 +5,5 @@ ext {
     alfresco_repo_version = '7.33.12'
     alfresco_dm_version = alfresco_61_dm_version
     alfresco_min_version = "6.1"
-    //Not setting alfresco_max_version here to make it easier to test on next Alfresco version (only for latest version)
+    alfresco_max_version = "6.1.99"
```
RVanhuysseXenit, line 8: Property still needs to be added to `apix-impl/61/src/main/config/module.properties`?

apix-impl/50/src/main/java/eu/xenit/apix/alfresco50/SpringConfiguration.java
WimCrols: Good question. I tested this change by running all integration test on all versions -> everything succeeded.
Do you think the workflow engine is not covered by those?

apix-impl/config/alfresco-global.properties
WimCrols: I see. I think that from your commits then?
Can you explain a bit more? What was the problem you saw, and where did you find this fix?

apix-integrationtests/build.gradle
WimCrols: Hm.. I don't think so, since all tests work and they all use the same version of DE which (I think) solves it.

apix-impl/61/overrides.gradle
WimCrols: Good catch.

apix-impl/50/src/main/java/eu/xenit/apix/alfresco50/SpringConfiguration.java
RVanhuysseXenit: I checked the tests and it looks like the default engine is sufficiently covert. The APS option is not, but it is as a whole unfinished and not a priority. Additionally, the comment in the `ApixSpringConfiguration` class makes me think we do not need it atm.

apix-impl/config/alfresco-global.properties
RVanhuysseXenit: Looking at it again, a functional change wont be necessary. 
The referenced interceptor was added in 6.2 (hence no impact on older versions). I remembered incorrectly that it was changed.
What it tries to address is a 'gap' in security where a user that is not allowed to change the content of a node, could still change the content url property, and so circumvent the restriction. It does this by straight up disallowing updates to the contenturl property. It was intended to have a whitelist property, but this is bugged into a not working state.
Since apix creates and changes nodes through the contenturl property, it needs this disabled.

RVanhuysseXenit (dismissed):

apix-impl/config/alfresco-global.properties
WimCrols: Tnx Robrecht. I'll add that.

RVanhuysseXenit (approved):

PR #93, opened 2020-10-07
ALFREDAPI-453: Make publish explicit in Jenkinsfile

WimCrols:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-453

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,6 +32,16 @@ def BuildVersionX(version) {
     sh "${gradleCommand} :apix-integrationtests:test-${version}:integrationTest"
 }
 
+def Publish(sonatypeUsername, sonatypePassword, gpgPassPhrase, projectName) {
+    def gradleCommand = "./gradlew --info --stacktrace "
+    sh "${gradleCommand} ${projectName}:publish " +
+            " -Ppublish_username=${sonatypeUsername} " +
+            " -Ppublish_password=${sonatypePassword} " +
+            " -PkeyId=DF8285F0 " +
+            " -Ppassword=${gpgPassPhrase} " +
+            " -PsecretKeyRingFile=/var/jenkins_home/secring.gpg "
+}
+
 node {
     def gradleCommand = "./gradlew --info --stacktrace "
     try {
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -68,12 +78,13 @@ node {
                 withCredentials([
                         usernamePassword(credentialsId: 'sonatype', passwordVariable: 'sonatypePassword', usernameVariable: 'sonatypeUsername'),
                         string(credentialsId: 'gpgpassphrase', variable: 'gpgPassPhrase')]) {
-                    sh "${gradleCommand} publish " +
-                            " -Ppublish_username=${sonatypeUsername} " +
-                            " -Ppublish_password=${sonatypePassword} " +
-                            " -PkeyId=DF8285F0 " +
-                            " -Ppassword=${gpgPassPhrase} " +
-                            " -PsecretKeyRingFile=/var/jenkins_home/secring.gpg "
+                    Publish(sonatypeUsername, sonatypePassword, gpgPassPhrase, "apix-interface")
```
vierbergenlars, line 81: Eh, you know you can ensure that you only apply the `maven-publish` plugin on projects that you want to publish, right?
Or at least run all those tasks in one gradle invocation.

vierbergenlars, line 81: Calling gradle so many times, every time to publish only one part might not work, because the artifacts may be published to different sonatype repositories.

Instead, you can either

- Run one gradle command with all the projects that need to be published
- Only apply the `maven-publish` plugin to the projects that should be published.

WimCrols: Both @koenvangeerteruy and me tried to apply `maven-publish ` selectively first, since it seemed the cleanest, but it didn't work. We ran into problems with the `afterEvaluate` blocks, IIRC.

I can go for the 1st solution though.

WimCrols:
Ok, I think I got it with (more) selectively applying `maven-publish` @vierbergenlars @koenvangeerteruy .
I started a Jenkins replay build where I overrode publishing step to happen even though this is not m̀aster`branch to be 100% sure: https://jenkins-2.xenit.eu/job/Xenit%20Github/job/alfred-api/job/ALFREDAPI-453-fixpublish/14/console

RVanhuysseXenit:
Seems good to me

RVanhuysseXenit (approved):

koenvangeerteruy (approved):

PR #97, opened 2020-10-21
Alfredapi 463

rschev:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-463

- [n/a] Does the PR comply to REST HTTP result codes policy outlined in the [developer guide](https://github.com/xenit-eu/alfred-api/blob/master/developer-documentation)?
- [n/a] Is error handling done through a method annotated with `@ExceptionHandler` in the webscript classes?


apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
```diff
@@ -438,4 +438,30 @@ public void invalidLong_onlyNode_doesNotThrowException() {
         searchService.query(searchQuery);
     }
 
+    @Test
+    public void testSearchForQuotationMarkProperty() {
+        final String testString = "For testing \"quotes\".";
+
+        transactionService.getRetryingTransactionHelper().doInTransaction(() -> {
```
RVanhuysseXenit, line 445: Can you not use the `baseTest.init()` method to set this up, at least partially?

RVanhuysseXenit (dismissed):

vierbergenlars (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 
 ### Fixed
+* [ALFREDAPI-463](https://xenitsupport.jira.com/browse/ALFREDAPI-463): Fix quotation marks in searches being improperly escaped
```
WimCrols, line 13: Like I said in our daily this morning: make it clear some searches can break here:
`Fix quotation marks in searches being improperly escaped. Search queries can now be escaped properly. E.g. \"Compas Format\" instead of \\\"Compas Format\\\" as was needed previously. `

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/FtsNodeVisitorTest.java
```diff
@@ -54,7 +54,7 @@ public void TestTermALL() {
     public void TestTermEscaped() {
         String t = "type";
         SearchSyntaxNode node = builder.term(t, "my\"Val").create();
-        Assert.assertEquals(t.toUpperCase() + ":\"my\"Val\"", toFts(node));
```
WimCrols, line 57: Please keep as many tests as possible out of integration test, which is **the** biggest bottleneck in APIX build time.
Please move this to the unit tests. The best solution would to move this entire class (I think that should be be possible without mocking).

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
```diff
@@ -438,4 +438,30 @@ public void invalidLong_onlyNode_doesNotThrowException() {
         searchService.query(searchQuery);
     }
 
+    @Test
+    public void testSearchForQuotationMarkProperty() {
```
WimCrols, line 442: What is being tested here? Syntax is already tested in unit test above. Integration with Solr is tested in other integration tests. So we are actually testing Solr itself here, which I think is not needed.

WimCrols (changes requested):

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/SearchServiceTest.java
rschev: The test confirms that you can actually find a document with a quotation mark in it. The previous test suite unit-tested the syntax, but due to the lack of an integration test didn't notice that the test was incorrect/insufficient.The unit test doesn't entirely prevent regressions, since an abstraction layer can easily be added that interprets backslashes an additional time before the query is actually executed. The integration test would catch that.

The query is transactional so it doesn't need to wait for Solr, meaning that unlike some other search tests, it doesn't increase the build time by several seconds.

WimCrols: You have a good argument (although it can be applied to (too) many tests, and at a certain point you have to start pruning) but it's ok for me to keep it in.

apix-impl/src/integration-test/java/eu/xenit/apix/tests/search/FtsNodeVisitorTest.java
```diff
@@ -1,4 +1,4 @@
-package eu.xenit.apix.tests.search;
+package java.eu.xenit.apix.tests.search;
```
WimCrols, line 1: java.eu.xenit? That doesn't sound right. I'll approve, but please check it.

WimCrols (dismissed):

RVanhuysseXenit (approved):

PR #99, opened 2020-10-27
ALFREDAPI-401 move integrationtest shared src and replace codehaus

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-401


todorinskiz (dismissed):

apix-integrationtests/build.gradle
```diff
@@ -13,7 +13,7 @@ subprojects {
     sourceSets {
         main {
             java {
-                srcDir '../../apix-impl/src/integration-test'
+                srcDir '../src/integration-test/java'
```
WimCrols, line 16: Did you check code navigation is still working in IntelliJ after this change? (This will require a reindexing first otherwise you don't see if it broke)

WimCrols (changes requested):

apix-integrationtests/build.gradle
vierbergenlars, line 16: It is a bit strange to put code from the `main` sourceSet here inside the `integration-test` folder? Can't we rename it to `main` and just drop the sourceset configuration then?

Messing with `runtimeClasspath` may still be necessary unfortunately.

vierbergenlars (changes requested):

apix-integrationtests/build.gradle
```diff
@@ -5,29 +5,42 @@ plugins {
...
 subprojects {
     apply from: "${project.projectDir}/overrides.gradle"
-    apply plugin: 'eu.xenit.de'
-    apply plugin: 'eu.xenit.alfresco'
 
     sourceSets {
         main {
             java {
-                srcDir '../../apix-impl/src/integration-test'
+//                srcDirs = ['src/main/java', '../src/main/java']
```
vierbergenlars, line 23: Remove this commented line

apix-integrationtests/build.gradle
```diff
@@ -5,29 +5,41 @@ plugins {
...
-            }
         }
     }
 
+    dependencies {
+        compile(project(":apix-integrationtests"))
+    }
+}
+
+allprojects {
```
vierbergenlars, line 34: Cross-configuration for all projects?
Did you want to only apply this to subprojects?
Usually configuration for all projects is only done in the root build.gradle

WimCrols: I'm not the author, but I think this is copied from the `apix-impl` project tree, where it exists to make IntelliJ code navigation work in both shared and version-specific folders. Whether it's needed here I did not test.

RVanhuysseXenit: Its required. We could shift it to subprojects only if we put the shared source in its own subproject?

vierbergenlars: Can we then please move `allprojects` configuration to the root build.gradle?
It should not be spread around different subprojects

WimCrols: No, these settings must only be apploied to this subtree.
To illustrate:
- root
  - apix-integrationtest
   - 50
   - 51
- other stuff
It must apply to apix-integrationtest, 50, 51, ... but **not** to root or other stuff.

vierbergenlars: Oh, I have always misunderstood what `allprojects` means. I just looked it up now.

I thought it was *all* projects in the build. But it appears to be this project + subprojects.

vierbergenlars (approved):

WimCrols (approved):

PR #101, opened 2020-12-11
NYCSANSUP-34 Rerun site component retrieval upon accessdenied.

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/NYCSANSUP-34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Fixed
 * [ALFREDAPI-463](https://xenitsupport.jira.com/browse/ALFREDAPI-463): Fix quotation marks in searches being improperly escaped. Search queries can now be escaped properly. E.g. \"Compas Format\" instead of \\\"Compas Format\\\" as was needed previously.
 * [ALFREDAPI-466](https://xenitsupport.jira.com/browse/ALFREDAPI-466): Fix usage of the special `-me-` argument for the peopleAPI v1 & v2
+* [ALFREDAPI-472](https://xenitsupport.jira.com/browse/ALFREDAPI-472): Fix AccessDeniedException in sitesService (primarily from Alfresco Records Management)
 
 ### Deleted
 
--- a/apix-impl/src/main/java/eu/xenit/apix/alfresco/sites/SiteService.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/alfresco/sites/SiteService.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.alfresco.repo.security.permissions.AccessDeniedException;
 import org.alfresco.service.ServiceRegistry;
 import org.alfresco.service.cmr.site.SiteInfo;
 import org.slf4j.Logger;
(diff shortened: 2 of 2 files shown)


apix-impl/src/main/java/eu/xenit/apix/alfresco/sites/SiteService.java
```diff
@@ -43,26 +45,47 @@ public SiteService(ServiceRegistry serviceRegistry, ApixToAlfrescoConversion api
...
+                // that the user has at least read rights on the current site, we will retrieve it anyway for display
+                // purposes. If the user does indeed not have access to the component nodes, then they should be blocked
+                // when trying to access those specifically.
+                try {
+                    componentsMap = getSiteComponents(siteService, shortName);
+                } catch (AccessDeniedException accessDeniedException) {
+                    logger.debug("User {} does not have access to a site component for site {} according to exception."
+                            + " Overriding to return a complete list of all sites this user is member of.",
+                            userId, userSite, accessDeniedException);
+                    componentsMap = AuthenticationUtil.runAsSystem(() -> {
```
kpiot123, line 67: @RVanhuysseXenit I have my doubts here. If the complete list of sites is retrieved and afterwards you want to get some property from one of the sites on which you don't really have read permission, will there be another access denied exception? I haven't reproduced the issue myself. Have you?

RVanhuysseXenit: yes it will result in another access denied exception. If I understand the RM docs correctly, there is supposed to be a discrepancy between being able to see an item exists and being able to actually see it. For as far as I can reporduce it (eg by forcing the aspect at issue onto the documentlibrary through c4a) this seems to be intended behaviour. The test user can see the site and enter it, but they cannot retrieve info from the documentlibrary node and its children

kpiot123: @RVanhuysseXenit Then I would opt for leaving out the ones on which you don't have access due to Records Management. If the list of sites the user has access to is retrieved he/she probably wants to do something with that information. In that case an access denied exception would occur, which is confusing.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 ### Fixed
 * [ALFREDAPI-463](https://xenitsupport.jira.com/browse/ALFREDAPI-463): Fix quotation marks in searches being improperly escaped. Search queries can now be escaped properly. E.g. \"Compas Format\" instead of \\\"Compas Format\\\" as was needed previously.
 * [ALFREDAPI-466](https://xenitsupport.jira.com/browse/ALFREDAPI-466): Fix usage of the special `-me-` argument for the peopleAPI v1 & v2
+* [NYCSANSUP-34](https://xenitsupport.jira.com/browse/NYCSANSUP-34): Fix AccessDeniedException from alfresco-rm in sitesService
```
kpiot123, line 14: @RVanhuysseXenit Just 1 more comment. We actually need an ALFREDAPI ticket here that is related to the NYCSANSUP ticket.

WimCrols: Yes, please.
And while you are at it, could you write `Alfresco Records Management` in full for clarity?

apix-impl/src/main/java/eu/xenit/apix/alfresco/sites/SiteService.java
```diff
@@ -43,26 +45,44 @@ public SiteService(ServiceRegistry serviceRegistry, ApixToAlfrescoConversion api
...
-            componentsMap.put(WIKI_COMPONENT, wiki);
-            componentsMap.put(DISCUSSIONS_COMPONENT, discussions);
 
-            apixSites.add(new Site(nodeRef, shortName, title, description, isPublic, componentsMap));
+                NodeRef nodeRef = c.apix(userSite.getNodeRef());
+                String shortName = userSite.getShortName();
+                String title = userSite.getTitle();
+                String description = userSite.getDescription();
+                boolean isPublic = userSite.getIsPublic();
+                Map<String, NodeRef> componentsMap;
```
WimCrols, line 54: (non-blocking) If you want to avoid this declaration without assignment, you could also move the ` apixSites.add(..)` inside the try.

WimCrols (approved):

kpiot123 (approved):

PR #103, opened 2020-12-18
ALFREDAPI-470 add maxItems parameter to searchQueries

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-470


kerkhofsd:
Can you please describe the difference between `limit` and `maxItems` from the perspective in the API consumer? 
IMO having two restrictions parameters is making the API very confusing. 

Why exactly do we need those two as separate parameters? Because we still want to perform an unlimited search to make sure the total result count is correct? If that's the (only) case, I think we just need to figure out a different way to get the correct number of results from Solr.

WimCrols:
Closing this in favor of PR #104

PR #104, opened 2020-12-23
ALFREDAPI-470: Search limit & memory problems

WimCrols:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-470

### DIFF
--- a/apix-impl/src/main/java/eu/xenit/apix/alfresco/search/SearchService.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/alfresco/search/SearchService.java
@@ -12,7 +12,6 @@ import eu.xenit.apix.utils.java8.Optional;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
-import org.alfresco.service.cmr.repository.NodeRef;
 import org.alfresco.service.cmr.repository.StoreRef;
 import org.alfresco.service.cmr.search.QueryConsistency;
 import org.alfresco.service.cmr.search.ResultSet;
@@ -88,7 +87,7 @@ public class SearchService implements ISearchService {
         // TODO: correctly implement skip and limit, now just manually limiting
         // 5.x will add maxItems to searchParameters
         // 4.2 will bodge it by putting maxItems in a variable for use during results parsing
-        setSearchLimit(searchParameters, maxItems);
+        setSearchMaxItems(searchParameters, maxItems);
 
         // Skip count
         if (argSkipCount > 0) {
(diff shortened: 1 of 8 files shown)


apix-impl/src/main/java/eu/xenit/apix/alfresco/search/SearchService.java
```diff
@@ -150,17 +152,16 @@ protected SearchParameters buildSearchParameters(SearchQuery postQuery) {
         return searchParameters;
     }
 
-    // These 2 methods are overridden in the 4.2 impl
     protected void setSearchLimit(SearchParameters searchParameters, int max) {
```
RVanhuysseXenit, line 155: To improve clarity, can we change the getter and setter to getMaxItems and setMaxItems?

apix-impl/src/test/java/eu/xenit/apix/tests/search/SearchServiceUnitTest.java
```diff
@@ -121,6 +122,27 @@ public void testOrderBy_withMultivalueProperty_throwsIllegalArgumentException()
         apixSearchServiceMocked.query(query);
     }
 
+    @Test
+    public void testLimitDefault() {
```
RVanhuysseXenit, line 126: Same rename here

RVanhuysseXenit (changes requested):

apix-impl/src/test/java/eu/xenit/apix/tests/search/SearchServiceUnitTest.java
WimCrols: The parameter is actually called `limit` so I think the name should stay.

kerkhofsd:
If I'm not mistaking, we still don't have a test that makes sure the query to / response from Solr is limited? 

Given the amount of hassle this can cause, I think it is feasible to investigate if it's possible to add such a test? 
(When I investigated the issue, I added this: https://github.com/xenit-eu/alfred-api/commit/3fa1c32a29d68f4cf9dddabae1f11e7e24c3e5d1, what might or might not be a source of inspiration.)

RVanhuysseXenit (dismissed):

WimCrols:
> If I'm not mistaking, we still don't have a test that makes sure the query to / response from Solr is limited?
> 
> Given the amount of hassle this can cause, I think it is feasible to investigate if it's possible to add such a test?
> (When I investigated the issue, I added this: [3fa1c32](https://github.com/xenit-eu/alfred-api/commit/3fa1c32a29d68f4cf9dddabae1f11e7e24c3e5d1), what might or might not be a source of inspiration.)

@kerkhofsd Correct, because that would be a test for Solr, not for Alfred API in my opinion.

RVanhuysseXenit (approved):

PR #107, opened 2021-01-07
ALFREDAPI-445 add exception handler for fileExistsException

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-445

### DIFF
--- a/apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
+++ b/apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
@@ -1,0 +1,97 @@
+package eu.xenit.apix.rest.exceptions;
+
+import eu.xenit.apix.data.NodeRef;
+
+public class FileExistsException extends RuntimeException {
+
+    private NodeRef source;
+    private NodeRef parent;
+    private String name;
+
+    public FileExistsException() {
+    }
+
+    public FileExistsException(String message) {
+        super(message);
+    }
+
+    public FileExistsException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public FileExistsException(Throwable cause) {
+        super(cause);
+    }
+
+    public FileExistsException(String message, Throwable cause, boolean enableSuppression,
+            boolean writableStackTrace) {
+        super(message, cause, enableSuppression, writableStackTrace);
+    }
+
+    public FileExistsException(String message, NodeRef source, NodeRef parent, String name) {
+        super(message);
+        this.source = source;
+        this.parent = parent;
+        this.name = name;
+    }
+
+    public FileExistsException(String message, Throwable cause, NodeRef source, NodeRef parent, String name) {
+        super(message, cause);
+        this.source = source;
+        this.parent = parent;
+        this.name = name;
+    }
+
+    public FileExistsException(Throwable cause, NodeRef source, NodeRef parent, String name) {
+        super(cause);
+        this.source = source;
+        this.parent = parent;
+        this.name = name;
+    }
+
+    public FileExistsException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace,
+            NodeRef source, NodeRef parent, String name) {
+        super(message, cause, enableSuppression, writableStackTrace);
+        this.source = source;
+        this.parent = parent;
+        this.name = name;
+    }
+
+    public NodeRef getSource() {
+        return source;
+    }
+
+    public void setSource(NodeRef source) {
+        this.source = source;
+    }
+
+    public NodeRef getParent() {
+        return parent;
+    }
+
+    public void setParent(NodeRef parent) {
+        this.parent = parent;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public String toString() {
+        StringBuilder messagebuilder = new StringBuilder("A file with name '")
+                .append(getName())
+                .append("' already exists at ")
+                .append(getParent());
+        if (getSource() != null) {
+            messagebuilder.append("\n");
+            messagebuilder.append("The source node is ").append(getSource());
+        }
+        messagebuilder.append("\n");
+        messagebuilder.append(super.toString());
+        return messagebuilder.toString();
+    }
+}
--- a/apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
+++ b/apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
@@ -28,6 +28,12 @@ public class FileExistsException extends RuntimeException {
         super(message, cause, enableSuppression, writableStackTrace);
     }
 
+    public FileExistsException(NodeRef source, NodeRef parent, String name) {
+        this.source = source;
+        this.parent = parent;
+        this.name = name;
+    }
+
     public FileExistsException(String message, NodeRef source, NodeRef parent, String name) {
         super(message);
         this.source = source;
--- a/apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
+++ b/apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
@@ -1,6 +1,7 @@
 package eu.xenit.apix.rest.v1.internal;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.dynamicextensionsalfresco.webscripts.AnnotationWebScriptRequest;
 import com.github.dynamicextensionsalfresco.webscripts.arguments.ArgumentResolver;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
@@ -33,11 +34,21 @@ public class ApixArgumentResolver implements ArgumentResolver<Object, Annotation
                     argumentType, expectedParameterType));
         }*/
 
-        //TODO: inject apix object mapper?
-        ObjectMapper map = new ObjectMapper();
-
         try {
-            return map.readValue(request.getContent().getContent(), argumentType);
+            if (Exception.class.isAssignableFrom(argumentType)) {
+                if (request instanceof AnnotationWebScriptRequest) {
+                    logger.debug("Resolving to Exception from request object");
+                    AnnotationWebScriptRequest annotationWSRequest = (AnnotationWebScriptRequest) request;
+                    return annotationWSRequest.getThrownException();
+                } else {
+                    logger.error("Exception encounted but cannot handle based on request object");
+                    return null;
+                }
+            } else {
+                //TODO: inject apix object mapper?
+                ObjectMapper map = new ObjectMapper();
+                return map.readValue(request.getContent().getContent(), argumentType);
+            }
         } catch (IOException e) {
             logger.warn("Cannot convert webscript argument with type from package eu.xenit.apix to json", e);
             throw new RuntimeException("Cannot convert webscript argument (" + name + ") " +
--- a/apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
+++ b/apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
@@ -22,7 +22,7 @@ public class ApixArgumentResolver implements ArgumentResolver<Object, Annotation
     @Override
     public final boolean supports(final Class<?> parameterType, final Class<? extends Annotation> annotationType) {
         /* Determine if using Class.isAssignableFrom() breaks backwards compatibility. */
-        return parameterType.getCanonicalName().startsWith("eu.xenit.apix");
+        return parameterType.getCanonicalName().startsWith("eu.xenit.apix") && !Exception.class.isAssignableFrom(parameterType);
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -963,6 +963,14 @@ private void writeNotAuthorizedResponse(AccessDeniedException exception, WebScri
         writeJsonResponse(response, "Not authorised to execute this operation");
     }
 
+    @ExceptionHandler(FileExistsException.class)
+    private void writeFileExistsResponse(FileExistsException fileExistsException, WebScriptResponse response) throws IOException {
+        String message = String.format("A file with name %s already exists at %s", fileExistsException.getName(), fileExistsException.getParentNodeRef());
```
WimCrols, line 968: Do you know why it is get**parent**noderef? I know there is no accessor for it, but I assume the duplicate file should have its own noderef, no?

WimCrols (changes requested):
Check failing tests: https://jenkins-2.xenit.eu/blue/organizations/jenkins/Xenit%20Github%2Falfred-api/detail/ALFREDAPI-445/2/tests

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
RVanhuysseXenit: The called method handles all 3 functions (create, copy, move). The copy and move both have a `copyFrom` argument, which we could probably access and print. Ill try adding that. Obviously the create does not.
I suppose the idea behind the exception (from the alfresco code) is that it doesnt really matter where the node came from, so it only includes the parent? ill see if i can our code.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
```diff
@@ -33,11 +34,21 @@ public final Object resolveArgument(final Class<?> argumentType, final Annotatio
                     argumentType, expectedParameterType));
         }*/
 
-        //TODO: inject apix object mapper?
-        ObjectMapper map = new ObjectMapper();
-
         try {
-            return map.readValue(request.getContent().getContent(), argumentType);
+            if (Exception.class.isAssignableFrom(argumentType)) {
```
kerkhofsd, line 38: Can't you just not use this `ArgumentResolver` for `Exception`s? 

DE wil automatically resolve an exception parameter for `@ExceptionHandler` annotated methods.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
```diff
@@ -1,6 +1,7 @@
 package eu.xenit.apix.rest.v1.internal;
```
WimCrols, line 1: So I assume this whole class can be removed after https://xenitsupport.jira.com/browse/ALFREDAPI-476 ?

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -588,60 +588,68 @@ public void retrieveAncestors(@UriVariable String space, @UriVariable String sto
...
-                .nodeRefToNodeInfo(resultRef, this.fileFolderService, this.nodeService, this.permissionService);
+            NodeRef resultRef = new NodeRef(resultObject.toString());
 
-        writeJsonResponse(response, nodeInfo);
+            NodeInfo nodeInfo = this
+                    .nodeRefToNodeInfo(resultRef, this.fileFolderService, this.nodeService, this.permissionService);
+
+            writeJsonResponse(response, nodeInfo);
+        } catch (org.alfresco.service.cmr.model.FileExistsException fileExistsException) {
+            FileExistsException wrappingException = new FileExistsException(fileExistsException);
```
WimCrols, line 647: These assignments (648-650) should be part of the wrapping FileExistsException constructor. Otherwise every place where we use this, will have to copy-paste these same assignments.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/nodes/NodesWebscript1.java
```diff
@@ -963,6 +976,21 @@ private void writeNotAuthorizedResponse(AccessDeniedException exception, WebScri
         writeJsonResponse(response, "Not authorised to execute this operation");
     }
 
+    @ExceptionHandler(FileExistsException.class)
+    private void writeFileExistsResponse(FileExistsException fileExistsException, WebScriptResponse response) throws IOException {
+        StringBuilder messagebuilder = new StringBuilder("A file with name ")
```
WimCrols, line 981: Would you mind adding quotes (`'`) around `fileExistsException.name` to improve readability a little bit in the message?

apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
```diff
@@ -0,0 +1,54 @@
+package eu.xenit.apix.rest.exceptions;
+
+import eu.xenit.apix.data.NodeRef;
+
+public class FileExistsException extends RuntimeException {
```
WimCrols, line 5: (non-blocking) I would like to clean this class up a bit:
- make fields private
- remove unused methods (such as setters and unused ctors)

WimCrols (changes requested):

apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
RVanhuysseXenit: im not convinced removing the ctors is a good idea, seeing as exceptions follow a pretty standard pattern. Not having the expected ones is at best an annoyance, at worst an active encouragement not to use the exception.
I put this one in with an eye on possibly removing alfresco dependencies from the interface and rest module. That would imply that the apix services will have to also use this (and other) exceptions, Which in turn would lead to people having to interact with these exceptions directly. Ergo the need for the standard ctors.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
RVanhuysseXenit: that would be the goal of that ticket, yes, Though it's up to you and Thijs whether we actually go though with that.
@kerkhofsd I consider removing the resolver out of scope for this ticket. Unless you've got a strong argument for removing it now, just going with the modification seems most efficient to me.

kerkhofsd: I agree that removing the `ApixArgumentResolver` in favor of the `@RequestBody` functionality is a different scope.

kerkhofsd: Or let me correct: defining the scope of this ticket not up to me. 
I agree that removing the `ApixArgumentResolver` in favor of the `@RequestBody` is a considerable amount of work that can be handled separately.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
WimCrols: I can see your point.
Was non-blocking for me anyway, so I'll leave decision to you.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
WimCrols: Yes, out of scope

kerkhofsd: Not sure if you tagged me here because of my initial remark. => See additional comment there.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
kerkhofsd: Just to be clear: this comment is not related to moving from `ApixArgumentResolver` to `@RequestBody`. Exception arguments for methods annotated with `@ExceptionHandler(...)` do not require the `@RequestBody` annotation.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
```diff
@@ -0,0 +1,97 @@
...
+    public FileExistsException(Throwable cause) {
+        super(cause);
+    }
+
+    public FileExistsException(String message, Throwable cause, boolean enableSuppression,
+            boolean writableStackTrace) {
+        super(message, cause, enableSuppression, writableStackTrace);
+    }
+
+    public FileExistsException(String message, NodeRef source, NodeRef parent, String name) {
```
WimCrols, line 31: Thanks for the change, it's a lot of new constructors :smile: but the one I actually meant was a constructor that takes in an Alfresco FileExistsException. My understanding was that this exception exists only to wrap Alfresco exception, correct? So by adding that the existing repetition is removed:
```
public FileExistsException(org.alfresco.service.cmr.model.FileExistsException alfrescoException, NodeRef source) {
    this.source = source;
    this.parent = new NodeRef(alfrescoException.getParentNodeRef().toString()),
    this.name = alfrescoException.getName());
}
```
Anyway, this is a small remark and if I missed some good counterargument, let me know.

WimCrols (changes requested):

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
RVanhuysseXenit: I indeed end up in that resolver through DE, but it is not entirely clear to me how DE has selected that resolver. I suspect that the issue might be that it reuses the resolver it selected to handle the original incoming request.
It is also not clear to me how I'd configure DE to select another resolver (If my earlier hypothesis is correct, I dont think we can switch to another resolver for the existience of that request)

kerkhofsd: You can indicate if DE should use the resolver via the `supports(final Class<?> parameterType, final Class<? extends Annotation> annotationType)` method. 

AS-IS this resolver is applicable for the `FileExistsException` because the package of that class `.startsWith("eu.xenit.apix")`

My suggestion is to adapt the  `supports(...)` method so the `ApixArgumentResolver` is only used for model classes or at least not for Exceptions.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
RVanhuysseXenit: Ah I see.
Adding the original exception is considered best practice for maintaining the stacktrace to the source of the exception.
The duplicated fields are a coincidence of the assumption that an apix client would not know the alfresco librrary perse, and thus wouldnt be able to access the original fields.
But given that assumption there is also a point to ending the stacktrace at the apix level -and thus not passing on the original exception.

RVanhuysseXenit: Ah and not passing the alfresco Exception explicitly maintains the decoupling between apix and alfresco. So indeed one last constructor with just the source, parent and name fields would do the above improvement too.

tbh I first had the fields on the apix exception public to avoid the wildgrow of constructors, but if we're going by the books, they should be there.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
RVanhuysseXenit: aah, I see.
I misunderstood and changed the wrong method. ill give it a try with changing `supports`

apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
RVanhuysseXenit: Made an additional consideration, I first put the apix exception in the rest module, but I think it'd be more appropriate if the custom exceptions are stored in the interface module.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/v1/internal/ApixArgumentResolver.java
```diff
@@ -1,6 +1,7 @@
...
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.dynamicextensionsalfresco.webscripts.AnnotationWebScriptRequest;
```
kerkhofsd, line 4: Unnecessary import?

kerkhofsd (dismissed):
Only reviewed `ApixArgumentResolver`.

apix-rest-v1/src/main/java/eu/xenit/apix/rest/exceptions/FileExistsException.java
WimCrols: > Ah and not passing the alfresco Exception explicitly maintains the decoupling between apix and alfresco. 

True but balance it with the fact that it will always be used like that leading to the code duplication I'm trying to avoid here.
Also: it's also called **FileExistsException** so the coupling to the Alfresco exception is clear :smile: 

> I think it'd be more appropriate if the custom exceptions are stored in the interface module.

Agreed.

WimCrols (dismissed):

WimCrols (approved):

PR #108, opened 2021-01-11
Alfredapi 296

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-296

Pr will add a list of qnames for all aspects that are mandatory for the requested type or aspect definition.

### DIFF
--- a/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
@@ -9,6 +9,7 @@ import eu.xenit.apix.dictionary.aspects.IAspectService;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.alfresco.service.cmr.dictionary.DictionaryService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -46,6 +47,10 @@ public class AspectService implements IAspectService {
             properties.add(c.apix((qName)));
         }
         ret.setProperties(properties);
+        List<QName> mandatoryAspects = aspectDef.getDefaultAspects(true).stream()
+                .map(aspectDefinition -> c.apix(aspectDefinition.getName()))
+                .collect(Collectors.toList());
+        ret.setMandatoryAspects(mandatoryAspects);
         return ret;
 
     }
--- a/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
@@ -47,7 +47,7 @@ public class AspectService implements IAspectService {
             properties.add(c.apix((qName)));
         }
         ret.setProperties(properties);
-        List<QName> mandatoryAspects = aspectDef.getDefaultAspects(true).stream()
+        List<QName> mandatoryAspects = aspectDef.getDefaultAspects().stream()
                 .map(aspectDefinition -> c.apix(aspectDefinition.getName()))
                 .collect(Collectors.toList());
         ret.setMandatoryAspects(mandatoryAspects);
--- a/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/TypeService.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/TypeService.java
@@ -9,6 +9,7 @@ import eu.xenit.apix.dictionary.types.Types;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.alfresco.service.cmr.dictionary.DictionaryService;
 import org.alfresco.service.namespace.InvalidQNameException;
 import org.alfresco.service.namespace.NamespaceException;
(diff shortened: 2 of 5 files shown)


apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
```diff
@@ -46,6 +47,10 @@ private AspectDefinition GetAspectDefinition(org.alfresco.service.namespace.QNam
             properties.add(c.apix((qName)));
         }
         ret.setProperties(properties);
+        List<QName> mandatoryAspects = aspectDef.getDefaultAspects(true).stream()
```
vierbergenlars, line 50: This is including inherited aspects.
Is that on purpose?

I'm asking because there is also `getDefaultAspects()` (without boolean parameter) which also includes the inherited aspects.

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/TypeService.java
```diff
@@ -66,6 +67,10 @@ private TypeDefinition GetTypeDefinition(org.alfresco.service.namespace.QName qn
             properties.add(c.apix((qName)));
         }
         ret.setProperties(properties);
+        List<QName> mandatoryAspects = typeDef.getDefaultAspects(true).stream()
```
vierbergenlars, line 70: This is including inherited aspects.
Is that on purpose?

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
WimCrols: I don't have a strong opinion, but isn't it better to have a complete overview of aspects?

apix-integrationtests/src/main/java/eu/xenit/apix/rest/v1/tests/DictionaryTest.java
```diff
@@ -147,4 +150,24 @@ public void missingDocumentType_status_404() throws IOException {
         assertEquals(404, httpResponse.getStatusLine().getStatusCode());
     }
 
+    public void assertMandatoryAspects(JSONObject jsonObject, String expectedMandatoryAspect) {try {
```
WimCrols, line 153: Reformat

apix-integrationtests/src/main/java/eu/xenit/apix/rest/v1/tests/DictionaryTest.java
```diff
@@ -147,4 +150,24 @@ public void missingDocumentType_status_404() throws IOException {
...
+        JSONArray jsonArray = jsonObject.getJSONArray("mandatoryAspects");
+        boolean aspectFound = false;
+        int jsonArrayLength = jsonArray.length();
+        int arrayIndex = 0;
+        while (!aspectFound & arrayIndex < jsonArrayLength) {
```
WimCrols, line 158: This is a bitwise `&` instead of logical `&&`

WimCrols: (non-blocking) By using a `for` you can roll l157 and 163 into this.

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
RVanhuysseXenit: Yes this was on purpose. 
The only argument pro local-aspects-only (local apects) I see is that when specifying a type, there is less elements to iterate/check.
But when creating a new node, or changing to a type that does not inherit from the current, it would be easy to lose/forget inherited aspects.
With that in mind, I suggest we go for completeness and pass the inherited aspects too.

as for the used method, we can indeed use the argumentless version. ill fix that.

apix-integrationtests/src/main/java/eu/xenit/apix/rest/v1/tests/DictionaryTest.java
```diff
@@ -147,4 +150,24 @@ public void missingDocumentType_status_404() throws IOException {
         assertEquals(404, httpResponse.getStatusLine().getStatusCode());
     }
 
+    public void assertMandatoryAspects(JSONObject jsonObject, String expectedMandatoryAspect) {try {
+        JSONArray jsonArray = jsonObject.getJSONArray("mandatoryAspects");
+        boolean aspectFound = false;
+        int jsonArrayLength = jsonArray.length();
```
WimCrols, line 156: (non-blocking) Why the temp var? jsonArray isn't modified in the loop.

apix-integrationtests/src/main/java/eu/xenit/apix/rest/v1/tests/DictionaryTest.java
```diff
@@ -147,4 +150,24 @@ public void missingDocumentType_status_404() throws IOException {
...
+        int arrayIndex = 0;
+        while (!aspectFound & arrayIndex < jsonArrayLength) {
+            String element = (String) jsonArray.get(arrayIndex);
```
WimCrols, line 159: Using https://stleary.github.io/JSON-java/org/json/JSONArray.html#getString-int- avoids the (String) cast.

WimCrols (changes requested):

apix-impl/src/main/java/eu/xenit/apix/alfresco/dictionary/AspectService.java
RVanhuysseXenit: So checking this again, and the `getDefaultAspects()` returns intherited apects by default, so I'd count that as an additional argument to continue that here

vierbergenlars (approved):

WimCrols (approved):

PR #115, opened 2021-07-07
Alfredapi 483 Remove support for alfresco 5.0 & 5.1

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-483

- [] Does this PR avoid breaking the API? 
    **Breaking change: dropped support for older versions**


apix-integrationtests/build.gradle
```diff
@@ -83,14 +85,16 @@ subprojects {
         dependsOn composeUpTask
         doFirst {
```
vierbergenlars, line 86: Instead of having it as a `doFirst` here, it is maybe better to have it as a `doLast` block on your `composeUpTask`?

Then it does not matter which task get executed first when doing the integration tests

apix-integrationtests/build.gradle
```diff
@@ -67,7 +67,9 @@ subprojects {
     alfrescoDynamicExtensions {
         repository {
             endpoint {
-                protocol = "http"
+                protocol = project.hasProperty('protocol') ? project.protocol : 'http'
```
vierbergenlars, line 70: It will always be http, which is also the default, so you can leave this configuration off.

apix-integrationtests/build.gradle
```diff
@@ -67,7 +67,9 @@ subprojects {
...
+                host = project.hasProperty('host') ? project.host : 'localhost'
+                port = project.hasProperty('port') ? project.port : 8080
                 serviceUrl = "/alfresco/service"
```
vierbergenlars, line 73: This is the default service url, no need to specify it either

apix-integrationtests/build.gradle
```diff
@@ -67,7 +67,9 @@ subprojects {
...
             }
             authentication {
```
vierbergenlars, line 75: This `admin:admin` is also the default, it does not have to be specified

WimCrols (approved):

PR #116, opened 2021-07-28
Alfredapi 486: Drop 6.0

WimCrols:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-486


apix-integrationtests/src/main/java/eu/xenit/apix/rest/v1/tests/BulkTest.java
```diff
@@ -52,6 +52,10 @@
     @Autowired
     ServiceRegistry serviceRegistry;
 
+    private String jsonObjectGetStringFromInt(JSONObject targetObject, String key) {
+        return String.valueOf(targetObject.getInt(key));
```
Ranecdev, line 56: You could change this to targetObject.optInt(...) so it returns null when the value is not an int

Ranecdev (approved):

apix-integrationtests/src/main/java/eu/xenit/apix/rest/v1/tests/BulkTest.java
WimCrols: Thanks, but if it's not a blocking comment, I rather keep it, since this has to work over different versions of the library and it took quite some builds to get it working for all.

PR #133, opened 2021-10-06
ALFREDAPI-494

kpiot123:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-494

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Added
 
 ### Changed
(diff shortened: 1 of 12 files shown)


todorinskiz:
@kpiot123 wouldn't it be handy at least to log the ignored properties (in debug?) 
```@JsonAnySetter
public void ignored(String name, Object value) {
    Log.debug(name + " : " + value);
}```

kpiot123:
> What’s the source/reason for the unknown properties in the request? Changing the api behaviour for a non-conforming input seems like the wrong solution to the problem.

Alfred Desktop sends out the "source" property. I can also fix it in the client but deploying a new client is for the customer often more work than deploying a new backend module.

PR #141, opened 2022-03-31
ALFREDAPI-498 handle version nodes appropriately for sourceAssociations

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-498

Improve handling of version nodes in the getSourceAssociations call: alfresco does not support this operation on these nodes. Change ensure calls result into an empty result rather than an exception, with the benefit of maintaining old established behaviour.

### DIFF
--- a/apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
@@ -411,14 +411,19 @@ public class NodeService implements INodeService {
 
     @Override
     public List<NodeAssociation> getSourceAssociations(eu.xenit.apix.data.NodeRef ref) {
-        return nodeService.getSourceAssocs(c.alfresco(ref), RegexQNamePattern.MATCH_ALL)
-                .stream()
-                .map(alfPeerAssoc ->
-                        new NodeAssociation(
-                                c.apix(alfPeerAssoc.getSourceRef()),
-                                ref,
-                                c.apix(alfPeerAssoc.getTypeQName())))
-                .collect(Collectors.toList());
+        // Versionstore does not support sourceAssocs. For version nodes, do not do call, add empty list to result
+        List<NodeAssociation> sourceAssocs = null;
+        if ("versionStore".equals(ref.getStoreRefProtocol())) {
+            sourceAssocs = nodeService.getSourceAssocs(c.alfresco(ref), RegexQNamePattern.MATCH_ALL)
+                    .stream()
+                    .map(alfPeerAssoc ->
+                            new NodeAssociation(
+                                    c.apix(alfPeerAssoc.getSourceRef()),
+                                    ref,
+                                    c.apix(alfPeerAssoc.getTypeQName())))
+                    .collect(Collectors.toList());
+        }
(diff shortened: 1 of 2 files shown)


WimCrols:
Can you extend the generated REST docs with a comment (e.g. ` "Version store nodes do not support source associations, hence these will not be returned" `) too?

apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
```diff
@@ -411,14 +411,19 @@ private boolean containsAspect(List<AspectDefinition> aspects, AspectDefinition
...
-                .stream()
-                .map(alfPeerAssoc ->
-                        new NodeAssociation(
-                                c.apix(alfPeerAssoc.getSourceRef()),
-                                ref,
-                                c.apix(alfPeerAssoc.getTypeQName())))
-                .collect(Collectors.toList());
+        // Versionstore does not support sourceAssocs. For version nodes, do not do call, add empty list to result
+        List<NodeAssociation> sourceAssocs = null;
+        if ("versionStore".equals(ref.getStoreRefProtocol())) {
```
WimCrols, line 416: Shouldn't this be **not** equals?

apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
```diff
@@ -411,14 +411,19 @@ private boolean containsAspect(List<AspectDefinition> aspects, AspectDefinition
...
+            sourceAssocs = nodeService.getSourceAssocs(c.alfresco(ref), RegexQNamePattern.MATCH_ALL)
+                    .stream()
+                    .map(alfPeerAssoc ->
+                            new NodeAssociation(
+                                    c.apix(alfPeerAssoc.getSourceRef()),
+                                    ref,
+                                    c.apix(alfPeerAssoc.getTypeQName())))
+                    .collect(Collectors.toList());
+        }
+        return sourceAssocs == null ? new ArrayList<>() : sourceAssocs;
```
WimCrols, line 426: IMO control flow is more readable like this:
```suggestion
        List<NodeAssociation> sourceAssocs = null;
        if ("versionStore".equals(ref.getStoreRefProtocol())) {
            return new ArrayList<>();
        }
        
        return nodeService.getSourceAssocs(c.alfresco(ref), RegexQNamePattern.MATCH_ALL)
                    .stream()
                    .map(alfPeerAssoc ->
                            new NodeAssociation(
                                    c.apix(alfPeerAssoc.getSourceRef()),
                                    ref,
                                    c.apix(alfPeerAssoc.getTypeQName())))
                    .collect(Collectors.toList());
```

WimCrols (changes requested):

apix-impl/src/main/java/eu/xenit/apix/alfresco/metadata/NodeService.java
WimCrols: The red line in my suggestion is a bit misleading, sorry. My suggestion is intended to replace the entire new code block.

WimCrols (approved):
Approving

PR #151, opened 2023-01-19
Alfredapi-509 GHA CI

WimCrols:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-509


.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Login to Docker
+        uses: docker/login-action@v2
+        with:
+          registry: private.docker.xenit.eu
+          username: ${{ secrets.CLOUDSMITH_USER }}
+          password: ${{ secrets.CLOUDSMITH_APIKEY }}
+      - name: Build interface  # Execute before integration testing to catch errors early
+        uses: gradle/gradle-build-action@v2.3.0
+        env:
+          XENIT_ARTIFACTORY_USERNAME: ${{ secrets.XENIT_ARTIFACTORY_USERNAME }}
```
RVanhuysseXenit, line 26: Why still use artifactory if we overwrite the commons-lang dep anyways?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+          XENIT_ARTIFACTORY_PASSWORD: ${{ secrets.XENIT_ARTIFACTORY_PASSWORD }}
+        with:
+          arguments: >-
+            --info
+            :apix-impl:apix-impl-${{ matrix.alfresco_version }}:test
+            :apix-integrationtests:test-${{ matrix.alfresco_version }}:integrationTest
+      - name: Upload Artifact
+        if: always()
```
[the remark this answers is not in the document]
WimCrols: Good question. It was either the choice between: A `success() || failure()` or B `always()` .
A (besides looking confusing) does not upload the report in case of cancellation of the build, which you do when the build hangs, so I went for B.
In the end, it seems it does not matter, because when cancelling or failing integration tests, only the unit test report is uploaded. Still, small preference for B.

.github/workflows/ci.yml
WimCrols: I think we still use other artifacts... I'll try without.

WimCrols: For now we still use acs-packaging that we mirror. We could switch to Alfresco nexus, but I would like to leave that our of scope for this one for now.

RVanhuysseXenit (approved):

PR #174, opened 2023-06-15
ALFREDAPI-521_release [Update] ci BRANCH_NAME

codingBenVdS:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-521

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,9 +57,10 @@ jobs:
         uses: mikepenz/action-junit-report@v3.7.6
         if: success() || failure()
         with:
-          check_name: 'test_reports'
+          check_name: "test-reports-${{ matrix.alfresco_version }}"
           report_paths: '**/build/test-results/**/TEST-*.xml'
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          detailed_summary: true
 
       - name: Publish
         if: ${{ startsWith(github.ref, 'refs/heads/master') || startsWith(github.ref, 'refs/heads/release') }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,14 @@ jobs:
           name: test-result
           path: /home/runner/work/**/build/reports
           retention-days: 2
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v3.7.6
+        if: success() || failure()
+        with:
+          check_name: 'test_reports'
+          report_paths: '**/build/test-results/**/TEST-*.xml'
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Publish
         if: ${{ startsWith(github.ref, 'refs/heads/master') || startsWith(github.ref, 'refs/heads/release') }}
         uses: gradle/gradle-build-action@v2.3.0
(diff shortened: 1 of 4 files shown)


WimCrols (dismissed):
Code looks good. Build failed... Do you already have a view on which tests fail?

.github/workflows/ci.yml
```diff
@@ -53,6 +53,14 @@ jobs:
           name: test-result
           path: /home/runner/work/**/build/reports
           retention-days: 2
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v3.7.6
+        if: success() || failure()
```
Brobrechts, line 58: if success or failure, is there another option or is this everything?

( success() == !failure()) ?

Brobrechts (commented):
I checked the report and it looks good,
There's a subsection per alfresco version, which is nice

But is this how we need to review it? We cannot browse it ad-hoc in GHA?

.github/workflows/ci.yml
WimCrols: @Brobrechts I had the same question. Turns the answer is: yes, there is something else: `canceled()`
There is even `always()`, but in their own docs they recommend using `sucess() || failure()` instead: https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

.github/workflows/ci.yml
```diff
@@ -53,6 +53,14 @@ jobs:
...
+        with:
+          check_name: "test-reports-${{ matrix.alfresco_version }}"
+          report_paths: '**/build/test-results/**/TEST-*.xml'
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
WimCrols, line 62: Maybe try `detailed_summary` https://github.com/mikepenz/action-junit-report?

apix-integrationtests/src/main/java/eu/xenit/apix/tests/workflow/methods/WorkflowService_SearchContextMyPooledTasks_Test.java
```diff
@@ -36,6 +37,11 @@ public Object execute() {
     }
 
     @Test
+    public void testDummy(){
+        System.out.println("test to resolve: NoRunableMethodExceptionExample: to Delete");
+        assertEquals(1,1);
+    }
+    @Ignore("Disabled for GHA build")
```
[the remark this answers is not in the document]
codingBenVdS: Looks like the race-condition is solved by placing a dummy test before.
Will test again by adding a delay at the start of the test.

.github/workflows/ci.yml
codingBenVdS: this adds the following when all tests succeed: 
image
This adds the following when test(s) fail:
image

codingBenVdS: comparing this to what we see when opening the "test_results":
image
i think we can let it out

WimCrols: I'm ok with both options. Thanks for trying it out.

WimCrols:
> I checked the report and it looks good, There's a subsection per alfresco version, which is nice
> 
> But is this how we need to review it? We cannot browse it ad-hoc in GHA?

You can read the report in GHA for example: https://github.com/xenit-eu/alfred-api/runs/14436069349 AFAIK that is the only place you can read it. 
What other place did you read it, @Brobrechts ?

apix-integrationtests/src/main/java/eu/xenit/apix/tests/workflow/methods/WorkflowService_SearchContextMyPooledTasks_Test.java
```diff
@@ -35,6 +35,11 @@ public Object execute() {
                 }, false, true);
     }
 
+    @Test
+    public void testDummy(){
```
WimCrols, line 39: Please format

WimCrols (approved):
One file that needs formatting, but I'll already approve.
Thanks for rebasing your commits :+1:

WimCrols (approved):

PR #177, opened 2023-06-20
Alfredapi 520 mvc branch fix

RVanhuysseXenit:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-520

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,15 @@
 # Alfred API - Changelog
 
-## 4.0.1 (yyyy-mm-dd)
+## 4.0.2 (yyyy-mm-dd)
 
 ### Added
 
 ### Changed
 
-
 ### Fixed
+* [ALFREDAPI-520](https://xenitsupport.jira.com/browse/ALFREDAPI-520): Enforce encoding on json responses to guarantee clean response in bulk webscript
 
 ### Removed
-
 
 ## 4.0.1 (unreleased)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,17 +1,16 @@
 # Alfred API - Changelog
 
-## 4.0.1 (yyyy-mm-dd)
+## 4.0.2 (yyyy-mm-dd)
```
WimCrols, line 3: I'm considering making version nr of MVC release 4.1.0 to set it apart from the old Alfred API. If you want you can already set it here.

WimCrols (changes requested):

CHANGELOG.md
RVanhuysseXenit: tbf, I'd consider changing it to 5.0 since the installation requirements are changing?

WimCrols: True, but it does not break any interface.
Anyway, let's keep that discussion out of scope for this PR.

WimCrols (approved):
:shipit:

RVanhuysseXenit:
@WimCrols I think you might have to override the merge, since it looks like the jenkins-build requirements are still applied to this pr?
image

WimCrols:
Hm, this branch probably still has a Jenkinsfile. Ok, I'll see what I can do.

WimCrols:
@RVanhuysseXenit I overrode it, but none of those status checks came from Jenkins, but from GHA actually, so I don't understand why we see them twice.

PR #202, opened 2024-03-19
Release 5.0.1

pvriel:
https://xenitsupport.jira.com/browse/ALFREDAPI-543


pvriel:
@WimCrols I can't review this PR myself, even though XEP-11 says this would be fine.
However, I do notice that the release branch is 9 commits AHEAD of master. No actual content though, but is this normal?

Brobrechts (approved):
As discussed, this is a formality

Only thing that remains to be checked is the 9 commits ahead of master with no code changes (diff shows nothing)

PR #218, opened 2024-06-14
[ALFREDAPI-554] make service beans available in main application context

pvriel:
https://xenitsupport.jira.com/browse/ALFREDAPI-554

Problem: Alfred API's service beans (e.g. ISearchService) are present in a separate child application context, making them unavailable for AMPs to be used.

This PR makes the beans of the `apix-impl` subproject available in the main application context of Alfresco, which Alfresco also uses when initializing beans of AMPs.

**Note to the reviewer:** I noticed that running the `test` Gradle task does not work. The following problems occur:
TypeServiceUnitTest:
	- testGetTypeDefinition_returnsNull_whenQnameInvalid
SetMetadataUnitTest:
	- testGeneralizeTypeWithCleanUpEnabled
	- testGeneralizeTypeWithCleanUpdisabled
SearchFacetServiceUnitTest:
	- assertThat_getFacetResults_returnIncludes_translationsForListOfValueConstraints
SearchServiceUnitTest:
	- all tests, except for testDefaultSearchQueryConsistencyIsTransactionalIfPossible
However, I noticed the same problem when running these tests on the master branch. These errors do not represent a new bug introduced by this PR.

### DIFF
--- a/apix-impl/src/main/java/eu/xenit/apix/SpringConfiguration.java
+++ b/apix-impl/src/main/java/eu/xenit/apix/SpringConfiguration.java
@@ -4,9 +4,7 @@ import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
-@ComponentScan(basePackageClasses = {
-        Version.class,
-})
+@ComponentScan(basePackages = {"eu.xenit.apix.alfresco"})
 public class SpringConfiguration {
 
 }
--- a/src/config/module-context.xml
+++ b/src/config/module-context.xml
@@ -10,4 +10,6 @@
             </list>
         </property>
     </bean>
+
+    <bean class="eu.xenit.apix.SpringConfiguration"/>
 </beans>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Alfred API - Changelog
 
+## 5.0.3 (2024-06-14)
+
+### Fixed
+
+* [ALFREDAPI-554](https://xenitsupport.jira.com/browse/ALFREDAPI-554): expose `apix-impl`
+beans in main application context (to be used by other AMPs)
+
 ## 5.0.2 (2024-05-14)
 
 ### Fixed
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ def static getVersionQualifier(String branch_name) {
 }
 
 ext {
-    versionWithoutQualifier = '5.0.2'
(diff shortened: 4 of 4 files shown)


apix-impl/src/main/java/eu/xenit/apix/SpringConfiguration.java
pvriel, line 1: This file was already present in the repo, but unused according to InteliJ (double-checked this by doing a full search + removing the class and using the `composeUp`-commands). However, when comparing this file to the structure of the repo, it looked like this file was originally added to accomplish the exact same goal this PR was created for. Therefore, I just updated this file & added it to the correct configuration file.

apix-impl/src/main/java/eu/xenit/apix/SpringConfiguration.java
```diff
@@ -4,9 +4,7 @@
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
-@ComponentScan(basePackageClasses = {
-        Version.class,
-})
+@ComponentScan(basePackages = {"eu.xenit.apix.alfresco"})
```
pvriel, line 7: `Version.class` does not work here, the Alfresco container crashes with the following message:

> 2024-06-14T07:05:32.724339994Z org.springframework.beans.factory.BeanDefinitionStoreException: Invalid bean definition with name 'schema.create.core' defined in URL [jar:file:/usr/local/tomcat/webapps/alfresco/WEB-INF/lib/alfresco-repository-21.20.jar!/alfresco/dbscripts/db-schema-context.xml]: Could not resolve placeholder 'db.script.dialect' in value "classpath:alfresco/dbscripts/create/${db.script.dialect}/AlfrescoCreate-RepoTables.sql"; nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'db.script.dialect' in value "classpath:alfresco/dbscripts/create/${db.script.dialect}/AlfrescoCreate-RepoTables.sql"

Using `basePackages = {"eu.xenit.apix.alfresco"}` was the best replacement I could come up with.

src/config/module-context.xml
```diff
@@ -10,4 +10,6 @@
             </list>
         </property>
     </bean>
+
+    <bean class="eu.xenit.apix.SpringConfiguration"/>
```
pvriel, line 14: Line responsible for making sure that Alfresco adds the beans from the `"eu.xenit.apix.alfresco"` (sub-)package(s) in the main application context.

WimCrols (approved):

PR #236, opened 2024-12-18
Backport 5.x/testing improvements

RVanhuysseXenit:
Testing improvements made for  https://xenitsupport.jira.com/browse/ALFREDAPI-566

- [N/A] Is [CHANGELOG.md](https://github.com/xenit-eu/alfred-api/blob/master/CHANGELOG.md) extended?
- [X] Are all Alfresco services wired via ServiceRegistry (and not per service)?
- [X] Does the PR comply to REST HTTP result codes policy outlined in the [user guide](https://docs.xenit.eu/alfred-api/user/rest-api/index.html#rest-http-result-codes)?
- [] Is usage of `this.` prefix avoided?

### DIFF
--- a/apix-docker/70/docker-compose.yml
+++ b/apix-docker/70/docker-compose.yml
@@ -2,7 +2,7 @@ version: '2'
 
 services:
   alfresco-core:
-    image: ${DOCKER_IMAGE:-private.docker.xenit.eu/alfresco-enterprise/alfresco-repository-enterprise:7.0}
+    image: ${DOCKER_IMAGE}
     ports:
       - "${DOCKER_IP}:8080:8080"
     volumes:
--- a/apix-docker/70/docker-compose.yml
+++ b/apix-docker/70/docker-compose.yml
@@ -2,7 +2,7 @@ version: '2'
 
 services:
   alfresco-core:
-    image: ${DOCKER_IMAGE}
+    image: ${DOCKER_IMAGE:-private.docker.xenit.eu/alfresco-enterprise/alfresco-repository-enterprise:7.0}
     ports:
       - "${DOCKER_IP}:8080:8080"
     volumes:
--- a/apix-integrationtests/alfresco/src/main/java/eu/xenit/apix/bootstrap/Bootstrap.java
+++ b/apix-integrationtests/alfresco/src/main/java/eu/xenit/apix/bootstrap/Bootstrap.java
@@ -1,0 +1,65 @@
+package eu.xenit.apix.bootstrap;
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+import org.alfresco.model.ContentModel;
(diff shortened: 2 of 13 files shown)


apix-docker/70/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '2'
 
 services:
   alfresco-core:
-    image: ${DOCKER_IMAGE}
+    image: ${DOCKER_IMAGE:-private.docker.xenit.eu/alfresco-enterprise/alfresco-repository-enterprise:7.0}
```
WimCrols, line 5: These defaults can lead to silent failures when the image is not built. I prefer to keep them out.

build.gradle
```diff
@@ -23,6 +23,7 @@ ext {
     jackson_version = '2.8.3'
     care4alfVersion = '2.3.0'
     http_version = '4.3.4' // Used by integration tests
+    ootbee_version = '1.2.2.0' // Used in docker image for testing
```
WimCrols, line 26: Since this is only used once, why not keep it locally, close to where it is used? It makes development easier since there is less switching needed.

apix-integrationtests/alfresco/src/main/java/eu/xenit/apix/bootstrap/Bootstrap.java
```diff
@@ -0,0 +1,75 @@
...
+        this.globalProperties = globalProperties;
+    }
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        if (Boolean.parseBoolean(globalProperties.getProperty(ENABLED_GLOBAL_PROPERTIES_KEY))) {
+            AuthenticationUtil.runAsSystem(
+                    () -> transactionHelper.doInTransaction(
+                            () -> {
+                                NodeRef companyHome = nodeLocator.getNode(CompanyHomeNodeLocator.NAME, null, null);
```
WimCrols, line 51: Be aware there is also https://github.com/xenit-eu/alfred-api/blob/master/alfred-api-integrationtests-client/alfresco/src/main/java/eu/xenit/alfred/api/tests/JavaApiBaseTest.java#L53 which may overlap? Non-blocking...

WimCrols (changes requested):

apix-integrationtests/alfresco/src/main/java/eu/xenit/apix/bootstrap/Bootstrap.java
WimCrols: Then I had have 2 remaining questions/remarks:

2. I understand better now, but it is a bit unintuitive for a class under package `integrationtesting`. I would suggest adding a small comment at least.
1. How is this well known test node then used exactly? By using Share manually or ...?

We can discuss IRL too when you have time.

PR #237, opened 2024-12-19
ALFREDAPI-565: Update swagger doc for /upload metadata

WimCrols:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-565
- [ ] Are all Alfresco services wired via ServiceRegistry (and not per service)?
- [ ] Does the PR comply to REST HTTP result codes policy outlined in the [user guide](https://docs.xenit.eu/alfred-api/user/rest-api/index.html#rest-http-result-codes)?

### DIFF
--- a/docs/swagger-ui/swagger.json
+++ b/docs/swagger-ui/swagger.json
@@ -620,7 +620,7 @@
           {
             "name": "metadata",
             "in": "formData",
-            "description": "Metadata for this file",
+            "description": "Metadata changes for this file. If used, the content type of this parameter (*not* of the request) needs to be set to *'application/json'*. E.g.: https://curl.se/docs/manpage.html#-F",
             "required": false,
             "type": "string"
           },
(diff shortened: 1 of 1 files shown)


docs/swagger-ui/swagger.json
```diff
@@ -636,11 +636,14 @@
           "200": {
             "description": "Success",
             "schema": {
-              "$ref": "#/definitions/NodeInfo"
+              "$ref": "#/definitions/MetadataChanges"
```
RVanhuysseXenit, line 639: This is incorrect, you'd need to change the `type` property in the `metadata` entry of `parameters`.
The returntype is unchanged.

RVanhuysseXenit (changes requested):

docs/swagger-ui/swagger.json
WimCrols: Oof, big mistake, indeed.

However, it seems to be impossible to use complex types (from https://swagger.io/docs/specification/v2_0/describing-parameters/#form-parameters):
> Form parameters are defined as in: formData. They can only be primitives (strings, numbers, booleans) or arrays of primitives (meaning you cannot use a $ref as the items value).

I will roll back this change, of course.

RVanhuysseXenit: Its not spec compliant, but the openapi-generator client gen doesnt fuss about it (with validation disabled).
It creates a method signature with `Object`, and then depending on the underlying http & json libs, it's might work or might not. The Spring libs seem to be able to handle it, the more generic clients not so much it seems.

RVanhuysseXenit (approved):

PR #251, opened 2025-09-09
Alfredapi-852: Migrate away from Cloudsmith

WimCrols:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-582

- [x] Are all Alfresco services wired via ServiceRegistry (and not per service), and is ServiceRegistry autowired with `@Qualifier("ServiceRegistry")`?
- [x] Does the PR comply to REST HTTP result codes policy outlined in the [user guide](https://docs.xenit.eu/alfred-api/user/rest-api/index.html#rest-http-result-codes)?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ ext {
     http_version = '4.5.13' // Used by integration tests
 }
 
-subprojects {
+allprojects {
     apply plugin: 'java'
 
     group = 'eu.xenit.alfred.api'


build.gradle
```diff
@@ -21,7 +22,7 @@ ext {
     http_version = '4.5.13' // Used by integration tests
 }
 
-subprojects {
+allprojects {
```
WimCrols, line 25: This is done because the Xenit conventions gradle plugins need a **group** and **version** set on root-level.

publish.gradle
```diff
@@ -30,24 +31,11 @@ publishing {
             }
         }
     }
-
-    repositories {
```
Brobrechts, line 34: We default to something now that we can remove the repo definition?

Brobrechts (approved):
Looks good, just minor comments

publish.gradle
WimCrols: Yes, from convention (oss) plugin

PR #252, opened 2025-09-11
Test with ElasticSearch

WimCrols:
See https://xenitsupport.jira.com/browse/ALFREDAPI-581

This PR served first as an exploration of how much ElasticSearch was compatible with Alfred API. Since with only improvements to _test_ code (not actual code) ES would be almost completely supported, this was also added.

Finally some documentation was added and updated.

### DIFF
--- a/alfred-api-integrationtests-server/src/main/java/eu/xenit/alfred/api/util/SolrTestHelperImpl.java
+++ b/alfred-api-integrationtests-server/src/main/java/eu/xenit/alfred/api/util/SolrTestHelperImpl.java
@@ -107,16 +107,15 @@ public class SolrTestHelperImpl implements SolrTestHelper {
     private void waitForCompletion(Supplier<Boolean> hasCompleted) throws InterruptedException {
         final int pause_in_seconds = 5;
         final int max_attempts = 20;
-        System.out.print("Waiting " + pause_in_seconds + " seconds for Solr to index content");
+        logger.info("Waiting " + pause_in_seconds + " seconds for Solr to index content");
         for (int i = 0; i < max_attempts; i++) {
             if (hasCompleted.get()) {
                 return;
             }
             for (int j = 0; j < pause_in_seconds; j++) {
-                System.out.print("..." + ((i * pause_in_seconds) + j));
+                logger.info("..." + ((i * pause_in_seconds) + j));
                 Thread.sleep(1000);
             }
-            System.out.println();
         }
     }
 


alfred-api-docker/234-elasticsearch/docker-compose.yml
```diff
@@ -0,0 +1,166 @@
...
+    image: ${DOCKER_IMAGE:-image_not_set}
+    ports:
+      - "8080:8080"
+      - "4578:4578" # coderunner-server
+    volumes:
+      - alfresco:/opt/alfresco/alf_data
+    restart: unless-stopped
+    environment:
+      - TERM=xterm
+      - TOMCAT_ALLOW_CASUAL_MULTIPART_PARSING=false
```
Brobrechts: - TERM=xterm
      - TOMCAT_ALLOW_CASUAL_MULTIPART_PARSING=false

Can you explain these environment settings?
Are these best practice?
Should we be applying them elsewhere?

alfred-api-docker/234-elasticsearch/docker-compose.yml
```diff
@@ -0,0 +1,166 @@
...
+  kibana:
+    image: kibana:7.17.28
+    mem_limit: 256m
+    ports:
+      - "5601:5601"
+    depends_on:
+      - elasticsearch
+
+  elasticsearch-live-indexing:
+    image: quay.io/alfresco/alfresco-elasticsearch-live-indexing:5.2.0
```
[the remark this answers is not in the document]
WimCrols: Ah, this was not intended to be committed. Due to a misunderstanding with Hyland support, I upgraded this. I have to revert this.

alfred-api-docker/234-elasticsearch/docker-compose.yml
```diff
@@ -0,0 +1,166 @@
...
+    volumes:
+      - alfresco:/opt/alfresco/alf_data
+    restart: unless-stopped
+    environment:
+      - TERM=xterm
+      - TOMCAT_ALLOW_CASUAL_MULTIPART_PARSING=false
+      - GLOBAL_messaging.broker.url=failover:(nio://activemq:61616)?timeout=3000&jms.useCompression=true
+      - GLOBAL_transform.service.enabled=true
+      - GLOBAL_transform.service.url=http://transform-router:8095/
+      - GLOBAL_localTransform.core-aio.url=http://transform-core-aio:8090/
```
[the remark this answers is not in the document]
WimCrols: You need local if you use Share, independent whether you use ATS.

alfred-api-integrationtests-client/alfresco/src/main/java/eu/xenit/alfred/api/tests/JavaApiBaseTest.java
```diff
@@ -32,22 +31,17 @@ public abstract class JavaApiBaseTest extends BaseApplicationContextTest {
 
 
     protected NodeRef getNodeAtPath(String path) {
```
Brobrechts: This protected function replaces all the getNodeAtPath functions that were copied between the child classes.
(DRY, it's good this is centralized here or in a utils class.)

I cannot help but notice this function is slightly different from the removed ones.

It looks better but is it functionally the same?
One does select nodes the other does a query?

alfred-api-integrationtests-client/alfresco/src/main/java/eu/xenit/alfred/api/tests/search/SearchServiceTest.java
```diff
@@ -378,7 +375,7 @@ public void invalidLong_onlyNode_doesNotThrowException() {
         SearchSyntaxNode node =
                 new QueryBuilder()
                         .startOr()
-                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", LONG_MAX_VALUE_PLUS_ONE)
+                        .property("{http://www.alfresco.org/model/system/1.0}node-dbid", LONG_MAX_VALUE)
```
Brobrechts: This seems like something that was built to overflow the max_long, now it's not used? Or the overflow already occurs on LONG_MAX_VAL?

alfred-api-integrationtests-server/src/main/java/eu/xenit/alfred/api/util/SolrTestHelperImpl.java
```diff
@@ -50,10 +49,15 @@ public SolrTestHelperImpl(
...
     public boolean areTransactionsSynced() {
-        long alfTransaction = alfrescoServerInfo.getAlfTransactionIdDAO();
-        long solrTransaction = solrAdminClient.getLastTransactionId();
-        logger.debug("alf transaction: {}, solr transaction: {}", alfTransaction, solrTransaction);
-        return alfTransaction <= solrTransaction;
+        try {
+            long alfTransaction = alfrescoServerInfo.getAlfTransactionIdDAO();
+            long solrTransaction = solrAdminClient.getLastTransactionId();
+            logger.debug("alf transaction: {}, solr transaction: {}", alfTransaction, solrTransaction);
+            return alfTransaction <= solrTransaction;
```
Brobrechts: Just. out of curiosity, since this is very disorienting.

should the statement not be alfTransaction == solrTransaction

alf transaction could never be lower than the solr transaction? because the solr transaction is matched on the  alfresco transaction? Unless alfresco was rolled back and it's in some kind of weird bizarre corrupt state?

It's a very confusing statement, but yes, it should work

alfred-api-integrationtests-client/alfresco/src/main/java/eu/xenit/alfred/api/tests/search/SearchServiceFacetsTest.java
```diff
@@ -21,24 +18,48 @@
...
 
     @Before
-    public void Setup() {
+    public void setup() {
+        facetService = getBean(SolrFacetService.class);
+
+        // Only execute these tests if we are using Solr.
+        // Bucketed facets don't work OOTB in the same way on ElasticSearch as Solr
+        String searchSubsystem = getBean(SearchService.class).getSearchSubsystem();
+        if (searchSubsystem.equals("elasticsearch")) {
```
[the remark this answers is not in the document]
WimCrols: Although I'm a huge fan of enums of strings, I choose not to use enums here because the Alfresco global property itself is not an enum.

alfred-api-integrationtests-server/src/main/java/eu/xenit/alfred/api/util/SolrTestHelperImpl.java
WimCrols: I only indented this, so if you don't mind I will keep changes it out of scope to prevent regression.

PR #268, opened 2026-06-29
Alfredapi 586 add 25.3 support

Brobrechts:
Fixes https://xenitsupport.jira.com/browse/ALFREDAPI-586

- [ X ] Is [CHANGELOG.md](https://github.com/xenit-eu/alfred-api/blob/master/CHANGELOG.md) extended?
- [ X ] Does this PR avoid breaking the API? 
- [ X ] Are all Alfresco services wired via ServiceRegistry (and not per service), and is ServiceRegistry autowired with `@Qualifier("ServiceRegistry")`?
- [ X ] Does the PR comply to REST HTTP result codes policy outlined in the [user guide](https://docs.xenit.eu/alfred-api/user/rest-api/index.html#rest-http-result-codes)?
- [ X ] Is error handling done through a method annotated with `@ExceptionHandler` in the webscript classes?
- [ X ] Does the PR follow our [coding styleguide and other active procedures](https://xenitsupport.jira.com/wiki/spaces/XEN/pages/624558081/XeniT+Enhancement+Proposals+XEP)?
- [ X ] Is usage of `this.` prefix avoided?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,11 @@
 ## 6.1.3 (unreleased)
 
 ### Added
+* [ALFREDAPI-586](https://xenitsupport.jira.com/browse/ALFREDAPI-586): Support Alfresco V25.3
 
 ### Fixed
 * [ALFREDAPI-587](https://xenitsupport.jira.com/browse/ALFREDAPI-587): Remove non-existant Swagger endpoints from documentation.
+
 
 ## 6.1.2 (2025-11-06)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
   test-impl:
     strategy:
       matrix:
-        alfresco_version: [ 231, 232, 233, 234, 251 ]
+        alfresco_version: [ 231, 232, 233, 234, 251, 253 ]
       fail-fast: false
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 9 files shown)


WimCrols:
Can I review this, seeing it is in draft?

CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
 # Alfred API - Changelog
 
-## 6.1.3 (unreleased)
+## 6.1.4 (unreleased)
 
 ### Added
 
 ### Fixed
 
+## 6.1.3 (2025-11-x)
```
WimCrols: Normally release dates are only set during a release. Then after release, a new `unreleased` chapter is added. The goal is that every release has clean, readable release notes.

WimCrols: (non-blocking) Additionally, according to semver this is a correct bump, but supporting a new Alfresco version is kind of big, so personally I would bump minor. To discuss.

WimCrols:
Note to self: after merge, add 25.3 to required CI checks.

WimCrols (changes requested):

CHANGELOG.md
Brobrechts: It's only a minor alfresco upgrade, so I would say it's only rerunning the tests.

For 26x; perhaps yes we could do a major increase

Brobrechts: Isn't this a chicken or the egg problem?
If you approve the PR, I will update this, you will need to re-approve, etc?

Perhaps  we should consider just using the github releases flow. we can even build per-releases there and have some more flexibility.

WimCrols: We have been doing it like this for year, no problem. Full procedure at https://xenitsupport.jira.com/wiki/spaces/APIX/pages/665190440/Release+procedure

WimCrols (approved):

